PR #4, opened 2020-12-24
feat: controller runtime implementation

smira:
This enhances Resource and State implementation, and build initial
Controller and Controller Runtime implementation. Code is reasonably
covered with tunit-tests.


smira:
we had a live review session on this PR, in general it looks good, it might need more changes as we get it integrated/used more, so we treat it as prototype/alpha version

PR #5, opened 2021-01-15
fix: preserve original YAML formatting in resource.Any

smira:
This helps to preserve server-side YAML formatting when resource spec
goes through YAML marshaling (e.g. in the API).

### DIFF
--- a/pkg/resource/any_test.go
+++ b/pkg/resource/any_test.go
@@ -5,9 +5,11 @@
 package resource_test
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v3"
 
 	"github.com/talos-systems/os-runtime/pkg/resource"
 )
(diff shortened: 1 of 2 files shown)


pkg/resource/any_test.go
```diff
@@ -30,6 +31,26 @@ func TestNewAnyFromProto(t *testing.T) {
...
+    type: type
+    id: aaa
+    version: 1
+    phase: running
+    finalizers:
+      - resource1
+      - resource2
+spec:
+    value: xyz
+    something: [a, b, c]
```
smira, line 54: this is not the default formatting, so this verifies that original flow-style is preserved here

pkg/resource/any_test.go
```diff
@@ -30,6 +31,26 @@ func TestNewAnyFromProto(t *testing.T) {
...
+	assert.Equal(t, map[string]interface{}{"something": []interface{}{"a", "b", "c"}, "value": "xyz"}, r.Value())
 	assert.Equal(t, "aaa", r.Metadata().ID())
+
+	enc, err := resource.MarshalYAML(r)
+	assert.NoError(t, err)
+
+	out, err := yaml.Marshal(enc)
+	assert.NoError(t, err)
+
+	assert.Equal(t, `metadata:
```
AlekSi, line 43:
```go
    assert.Equal(t, strings.TrimSpace(`
metadata:
  namespace: default
[…]
    something: [a, b, c]
    `) + "\n", string(out))
```

That allows one to keep YAML indented naturally in the test code.

Alternatively, that might be a separate assertion helper if there are more than a few similar comparisons.

AlekSi (approved):

Unix4ever (approved):

pkg/resource/any_test.go
smira: thanks, updated!

PR #10, opened 2021-01-21
feat: add support for exporting dependency graph

smira:
This graph shows relationship between resources and controllers, which
might be invaluable for debugging and understanding system dependencies.


andrewrynhard (dismissed):

pkg/controller/dependency.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// DependencyEdgeType is edge type in controller graph.
+type DependencyEdgeType int
+
+// Controller graph edge types.
+const (
+	EdgeManages DependencyEdgeType = iota
+	EdgeDependsHard
+	EdgeDependsWeak
```
andrewrynhard, line 21: Not for this PR but I missed this nit the other day. Strong and Weak go better together naturally. Hard and Soft likewise. I know this aligns with the current convention, but just my 2 cents for the future.

smira: thanks, let me update it for this PR until it spreads too far!

andrewrynhard (dismissed):

PR #13, opened 2021-03-02
chore: provide fmt.Stringer for EventType

smira:
Simple fix to allow easier printing of events.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -24,6 +24,10 @@ const (
 	Destroyed
 )
 
+func (eventType EventType) String() string {
+	return [...]string{"Created", "Updated", "Destroyed"}[eventType]
+}
+
 // Event is emitted when resource changes.
 type Event struct {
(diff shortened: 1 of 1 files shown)


andrewrynhard (approved):

pkg/state/state.go
```diff
@@ -24,6 +24,10 @@ const (
 	Destroyed
 )
 
+func (eventType EventType) String() string {
+	return [...]string{"Created", "Updated", "Destroyed"}[eventType]
```
AlekSi, line 28: Why not use `stringer` tool? https://pkg.go.dev/golang.org/x/tools/cmd/stringer

smira: never used, that, good point, if this requires changing build steps, might be too much work for this PR, but certainly something for the future

smira: https://github.com/talos-systems/kres/issues/53

PR #14, opened 2021-03-15
feat: use go-yaml fork and serialize spec as RawYAML objects

Unix4ever:
This should address incorrect comment indents as we're going to write
the spec as a raw YAML block.

Signed-off-by: Artem Chernyshev <artem.0xD2@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,10 @@ module github.com/talos-systems/os-runtime
 
 go 1.14
 
+// forked yaml that introduces RawYAML interface that can be used to provide YAML encoder bytes
+// which are then encoded as a valid YAML block with proper indentiation
+replace gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20210315173758-8fb30b8e5a5b
+
 require (
 	github.com/AlekSi/pointer v1.1.0
 	github.com/cenkalti/backoff/v4 v4.1.0
--- a/pkg/resource/any.go
+++ b/pkg/resource/any.go
@@ -18,12 +18,12 @@ type Any struct {
 
 type anySpec struct {
 	value interface{}
-	doc   yaml.Node
 	yaml  []byte
 }
 
-func (s anySpec) MarshalYAML() (interface{}, error) {
-	return s.doc.Content[0], nil
+// MarshalYAMLBytes implements RawYAML interface.
+func (s anySpec) MarshalYAMLBytes() ([]byte, error) {
+	return s.yaml, nil
 }
 
 // SpecProto is a protobuf interface of resource spec.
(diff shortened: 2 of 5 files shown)


pkg/resource/any_test.go
```diff
@@ -49,8 +49,8 @@ metadata:
     version: 1
     phase: running
     finalizers:
-      - resource1
-      - resource2
+        - resource1
```
smira, line 52: not sure why there's an indentation change here?

pkg/resource/any.go
```diff
@@ -18,12 +18,12 @@ type Any struct {
...
 type anySpec struct {
 	value interface{}
-	doc   yaml.Node
 	yaml  []byte
 }
 
-func (s anySpec) MarshalYAML() (interface{}, error) {
-	return s.doc.Content[0], nil
+// GetBytes implements RawYAML interface.
+func (s anySpec) GetBytes() ([]byte, error) {
```
smira, line 25: Should it be called better `MarshalYAMLBytes` or close to that?

pkg/resource/any_test.go
Unix4ever: We had an old version of go-yaml here.

pkg/resource/any.go
Unix4ever: yeah, I like that variant. Should be less chances to break the code for someone if they use that function name already.

Unix4ever: yeah, I like that variant. Should be less chances to break the code for someone if they use that function name already.

Unix4ever:
changes in yaml: https://github.com/Unix4ever/yaml/commit/8fb30b8e5a5b421cc57018e5c64689277140ebda

pkg/resource/any_test.go
Unix4ever:
```
commit ae27a744346343ea814bd6f3bdd41d8669b172d0
Author: Gustavo Niemeyer <gustavo@niemeyer.net>
Date:   Mon May 4 15:38:57 2020 +0100

    Make sequence indentation more usual
    
    Sequences used to be slightly different, and v3 made it slightly
    different in a slightly different way, and that's all slightly bad.
    All gone now. Sequences indent just like everything else.
```

The commit that changed that.

smira (approved):

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/talos-systems/os-runtime
 
 go 1.14
 
+replace gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20210315173758-8fb30b8e5a5b
```
AlekSi, line 5: I think it would be a good practice to add comments explaining all non-trivial go.mod directives: `replace`s, indirect dependencies, etc.

Unix4ever: good point, updated.

PR #18, opened 2021-03-29
feat: major rewrite of the os-runtime with new features

smira:
1. Add `owner` field to the resource metadata to track the controller
which owns the resource. State API now supports setting an owner and
enforces ownership check on object modifications. User-owned objects
(e.g. machine configuration in Talos) are supposed to have no owner.

2. Update controller API to have `Inputs` and `Outputs`: `Outputs` are
former `ManagedResources`, but controller can have now any number of
outputs, and outputs are defined only by resource type (namespace was
dropped). Outputs can be either exclusive (controller owns all the
resources of the type) or shared (multiple controllers might output
resources of the type, but every single instance is owned by a single
controller). Shared and exclusive outputs can't be mixed (if some
resource is an exclusive output of a controller, no other controller can
register it as a shared output). `Inputs` are just renamed
`Dependencies` without any changes. Initial set of inputs can be
provided now at the registration time.


api/v1alpha1/resource.proto
```diff
@@ -10,6 +10,7 @@ message Metadata {
     string type = 2;
     string id = 3;
     string version = 4;
+    string owner = 7;
```
andrewrynhard, line 13: nit: out of order.

api/v1alpha1/runtime.proto
```diff
@@ -38,16 +38,34 @@ service ControllerAdapter {
...
+message ControllerOutput {
     string type = 2;
+    ControllerOutputKind kind = 3;
+}
+
+// RegisterController RPC
+
+message RegisterControllerRequest {
+    string controller_name = 1;
+    repeated ControllerInput initial_inputs = 3;
```
andrewrynhard, line 67: My opinion, so take with a grain of salt: I think this can be simply `inputs`. To me it is clear that I can register with inputs and then update them after the fact. Calling them the same thing (just `inputs`) in each RPC solidifies that concept for me a bit more.

andrewrynhard (approved):
I focused on the protobufs mostly. Trusting the tests 🙂 .

api/v1alpha1/resource.proto
smira: I tried to preserve the tags to make sure it is backwards compatible, but I can break it if you're fine

api/v1alpha1/runtime.proto
smira: yes, I was torn on this one, I agree `Inputs/Outputs` sounds easier

api/v1alpha1/resource.proto
andrewrynhard: Oh I just meant that you can leave it as 7, but place it after six. But you do point out something I didn't consider. I guess the good thing about theses protobuf definitions for now are that we can change them with zero impact since it is completely self contained. In that case I would suggest "breaking things" even though in practice that doesn't mean much for us. Talos ships everything atomically.

smira: oh, this is not even consumed in Talos (yet), so doesn't matter, I was thinking about `os-engine`, but on other hand with other changes there's no backwards compatibility anyways, I'll get things updated

andrewrynhard (approved):

PR #22, opened 2021-04-19
feat: use zap logger instead of the system one

Unix4ever:
Zap logger gives more control over verbosity and log outputs.

Signed-off-by: Artem Chernyshev <artem.0xD2@gmail.com>


pkg/controller/runtime/runtime.go
```diff
@@ -25,7 +25,7 @@ type Runtime struct { //nolint: govet
 	depDB *dependency.Database
 
 	state  state.State
-	logger *log.Logger
+	logger *zap.Logger
```
souleb, line 28: What do you think about declaring a logger interface in logging package?

```go
type Logger interface {
  Log(msg string) error
}
```

We could then pass any implementation of it here. Would reduce the number of packages directly dependent on zap.

pkg/controller/protobuf/client/client.go
```diff
@@ -232,7 +235,7 @@ func (ctrlAdapter *controllerAdapter) run(ctx context.Context) {
 
 		interval := ctrlAdapter.backoff.NextBackOff()
 
-		logger.Printf("restarting controller in %s", interval)
+		logger.Debug(fmt.Sprintf("restarting controller in %s", interval))
```
AlekSi, line 238:
```suggestion
		logger.Sugar().Debugf("restarting controller in %s", interval)
```

The difference is that string formatting will not be done if debug level is disabled, saving a bit of CPU time

pkg/controller/runtime/runtime.go
Unix4ever: The problem with zap is that if you want to add context to your logs without using `zap.SugaredLogger`, you have to rely on `zap.Field`:

```go
logger.Info("hello there", zap.String("key", "value"))
```

What makes things worse is that `Field` is not an interface so you can't create an abstract interface for the logger this way.

Only if you create a wrapper instead... What makes thing ugly.

pkg/controller/protobuf/client/client.go
Unix4ever: yeah, that's an option, it just felt like calling `Sugar` for each log line is not how it's intended to be used.

AlekSi: It is, actually :) That's documented in README and their FAQ.

Another option could be to pass `*zap.SugaredLogger` instead, and call `Desugar()` where we need speed.

Unix4ever: :cool:

pkg/controller/runtime/runtime.go
souleb: I see.

Another suggestion. Not a standard though.

```go
func FromContext(ctx) *zap.Logger
```

We could then do this:

```go
func (ctrl *SumController) Run(ctx context.Context, r controller.Runtime) error {
  logger := logging.FromContext(ctx)
}
```

We could save on the number of parameters at least, if not direct dependencies

Unix4ever: Yeah... There's only a single problem with the context though: you can't just simply do:

```go
for _, key := range ctx.Keys() {
   ...
}
```

You can only extract values from it, only if your logger has the predefined list of keys.
Unless something has changed recently.

smira: https://pkg.go.dev/context

> Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.

I personally feel passing logger via context is not the best pattern. 

The idea here is that runtime can provide a scoped logger for each controller. 

At the same time if the controller is not running in the same process with this `runtime`, it uses its own logging and this change doesn't enforce anything on other implementation.

What we might really need from this implementation is to have a logger which is more configurable and flexible than stdlib `log.Logger`. If there are any better option than `zap`, we're all for it.

souleb: Going with the interface is the cleanest. We could take inspiration from Gokit. They have a zap implementatin here:
https://github.com/go-kit/kit/blob/master/log/zap/zap_sugar_logger.go

They use a simple interface

```go
type Logger interface {
	Log(keyvals ...interface{}) error
}
```

Unix4ever: Yeah, but I haven't used `SugaredLogger` for a reason. Here's the example:
https://github.com/talos-systems/theila/pull/13/files#diff-b21a01f7945651bbd6631a4fb9605f7c1e8d909c80f9727bd0d8d3c226e28046L33

There I migrated from sugared logger to a regular one.

It's very common that you define the same constants for log keys in many places, to avoid duplication you can use helper functions like:

```go
func Component(name string) zap.Field {
  return zap.String("component", name)
}
```

You _can_ do the same for SugaredLogger, but you'll have to return a key value pair:

```go
func Component(name) []interface{} {
  return "component", name
}
...

sugared.Infow("hello", Component("a")...)
```

And it works, until you get more than one field:

```go
// it's doable but starts to turn into a mess
sugared.Infow("hello", append(Component("b"), Error(err)...)...)
```

Or you'll have to do:

```go
sugared.Infow("hello", Component, "b", Error, err)
```

I would probably suggest to discuss that question on the next COSI meet.

pkg/controller/protobuf/client/client.go
```diff
@@ -232,7 +235,7 @@ func (ctrlAdapter *controllerAdapter) run(ctx context.Context) {
 
 		interval := ctrlAdapter.backoff.NextBackOff()
 
-		logger.Printf("restarting controller in %s", interval)
+		logger.Sugar().Debugf("restarting controller in %s", interval)
```
smira, line 238: this might be `Infof` at least as this is pretty important to debug why controller doesn't reconcile

smira (commented):
:+1: with some notes

pkg/controller/protobuf/client/client.go
Unix4ever: yeah. I just moved to debug everything that was producing too much noise, but if it's critical then let's move it to Info.

Unix4ever: updated

smira (approved):

smira (approved):

PR #26, opened 2021-04-23
chore: rename cosi-runtime to runtime

andrewrynhard:
Since the runtime now lives under the COSI project, we don't need "cosi"
in the name. We also discussed the general practice of not using the
project name (COSI) in the implementation. This change renames
"cosi-runtime" to "runtime" to align the project with this decision.

Signed-off-by: Andrew Rynhard <andrew@rynhard.io>

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -105,11 +105,11 @@ steps:
   depends_on:
   - unit-tests
 
-- name: cosi-runtime
+- name: runtime
   pull: always
   image: autonomy/build-container:latest
   commands:
-  - make cosi-runtime
+  - make runtime
   volumes:
   - name: outer-docker-socket
     path: /var/outer-run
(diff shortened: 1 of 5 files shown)


Unix4ever (approved):

stevendborrelli (approved):
LGTM pending successful CI

smira:
we need to do that via `kres`, as these changes will be overridden by the next build

andrewrynhard:
> we need to do that via `kres`, as these changes will be overridden by the next build

I ran `make rekres`. No diff.

andrewrynhard:
@smira see comment above. What am I missing?

smira:
> @smira see comment above. What am I missing?

I think it's myself missing the fact that you renamed the app binary, that should be enough for Kres to pick it up

smira (approved):

smira:
But still proper way might be to rename the folder and run Kres, should be easier. I was looking for the updated Kres comments in the build instructions as indicator )

PR #28, opened 2021-04-23
feat: add flag to listen on TCP

andrewrynhard:
Signed-off-by: Andrew Rynhard <andrew@rynhard.io>

### DIFF
--- a/api/v1alpha1/resource.pb.go
+++ b/api/v1alpha1/resource.pb.go
@@ -9,11 +9,12 @@
 package v1alpha1
 
 import (
+	reflect "reflect"
+	sync "sync"
+
 	proto "github.com/golang/protobuf/proto"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
-	reflect "reflect"
-	sync "sync"
 )
 
 const (
@@ -288,12 +289,15 @@ func file_v1alpha1_resource_proto_rawDescGZIP() []byte {
 	return file_v1alpha1_resource_proto_rawDescData
 }
 
-var file_v1alpha1_resource_proto_msgTypes = make([]protoimpl.MessageInfo, 3)
-var file_v1alpha1_resource_proto_goTypes = []interface{}{
-	(*Metadata)(nil), // 0: resource.Metadata
-	(*Spec)(nil),     // 1: resource.Spec
-	(*Resource)(nil), // 2: resource.Resource
-}
+var (
+	file_v1alpha1_resource_proto_msgTypes = make([]protoimpl.MessageInfo, 3)
+	file_v1alpha1_resource_proto_goTypes  = []interface{}{
+		(*Metadata)(nil), // 0: resource.Metadata
+		(*Spec)(nil),     // 1: resource.Spec
+		(*Resource)(nil), // 2: resource.Resource
+	}
+)
+
 var file_v1alpha1_resource_proto_depIdxs = []int32{
 	0, // 0: resource.Resource.metadata:type_name -> resource.Metadata
 	1, // 1: resource.Resource.spec:type_name -> resource.Spec
--- a/cmd/runtime/main.go
+++ b/cmd/runtime/main.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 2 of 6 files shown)


cmd/runtime/main.go
```diff
@@ -53,9 +56,28 @@ func run() error {
 		}
 	}()
 
-	l, err := net.Listen("unix", socketPath)
+	if err := os.Remove(socketPath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			log.Fatal(err)
+		}
+	}
```
andrewrynhard, line 63: Fixing this, `git` is broken for me right now. Will push a fix in a bit.

cmd/runtime/main.go
```diff
@@ -53,9 +56,28 @@ func run() error {
...
+		network = "unix"
+		address = socketPath
+
+		if err := os.Remove(socketPath); err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
+				log.Fatal(err)
+			}
+		}
+	} else {
+		network = "tcp"
```
invidian, line 74: Can't this be default? Then you can skip `else` block completely :)

cmd/runtime/main.go
```diff
@@ -26,10 +27,12 @@ import (
 	"github.com/cosi-project/runtime/pkg/state/protobuf/server"
 )
 
+var addressAndPort string
```
invidian, line 30: Must it be a global variable?

cmd/runtime/main.go
```diff
@@ -53,9 +56,28 @@ func run() error {
...
+		address string
+	)
+
+	if socketPath == "" {
+		network = "unix"
+		address = socketPath
+
+		if err := os.Remove(socketPath); err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
+				log.Fatal(err)
```
invidian, line 70: I think you should return error here instead of panicking.

invidian (commented):
Some suggestions :)

cmd/runtime/main.go
andrewrynhard: Whoops

cmd/runtime/main.go
andrewrynhard: Was already the pattern used. I just went with it.

andrewrynhard: Happy to address in a follow up, unless we feel it should be addressed in this PR.

andrewrynhard:
`make fmt` touched generated files. You can ignore.

api/v1alpha1/resource.pb.go
```diff
@@ -9,11 +9,12 @@
 package v1alpha1
 
 import (
+	reflect "reflect"
+	sync "sync"
+
 	proto "github.com/golang/protobuf/proto"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
-	reflect "reflect"
```
smira, line 15: not sure why these changes popped up?

cmd/runtime/main.go
```diff
@@ -26,10 +27,14 @@ import (
...
 
-var socketPath string
+var (
+	addressAndPort string
+	socketPath     string
+)
 
 func main() {
-	flag.StringVar(&socketPath, "socket-path", "/var/run/cosi-runtime.sock", "path to the UNIX socket to listen on")
+	flag.StringVar(&socketPath, "socket-path", "", "path to the UNIX socket to listen on")
```
smira, line 36: this changes default behavior by making runtime listen on TCP by default vs. file socket

I think having it listening on `*:50000` by default is not a good default

andrewrynhard: Default to unix socket? Default to `127.0.0.1`? Either way is fine by me.

smira: I would keep it as file socket, not saying it's more secure than localhost, but at least shows the intention better as the API doesn't have any auth at the moment.

cmd/runtime/main.go
```diff
@@ -26,10 +27,14 @@ import (
...
 
-var socketPath string
+var (
+	addressAndPort string
+	socketPath     string
+)
 
 func main() {
-	flag.StringVar(&socketPath, "socket-path", "/var/run/cosi-runtime.sock", "path to the UNIX socket to listen on")
+	flag.StringVar(&socketPath, "socket-path", "/system/runtime.sock", "path to the UNIX socket to listen on")
```
andrewrynhard, line 36: Updated to align with the `engine`'s assumptions.

smira (approved):

PR #35, opened 2021-05-24
feat: notify controllers when outputs are ready to be destroyed

### LINKED ISSUE #34: notify controller on outputs being ready for destroy
When controller output enters teardDown phase and has no finalizers, controller should be notified to run reconcile loop

smira:
Controllers should use two-phased approach when destroying their
outputs:

* first, resource is marked as being torn down (`Teardown()`), this also
prevents adding new finalizers
* second, when finalizers are removed by dependent controllers,
controller can finally completely remove its out with `Destroy()`

This problem is that controller should wait for the finalizers to be
empty between phases one and two. Waiting is possible, but controller
might not want to block other reconcilation activities while waiting for
the output to be ready to be destroyed.

This PR addresses this issue by allowing to declare controller output as
an input of special kind `InputDestroyReady` which means controller is
only interested to be notified when output is ready to be destroyed.

Fixes #34


api/v1alpha1/runtime.proto
```diff
@@ -127,6 +127,16 @@ enum ControllerInputKind {
...
+    // DestroyReady inputs provide notifications to the controllers when some resource
+    // is ready to be destroyed (all finalizers are removed).
+
+    // Usually outputs are declared as DestroyReady inputs to trigger reconcile loop
+    // when resource is ready to be destroy.
+    //
+    // Controller will be notified when the resource enters "teardown" phase and has no
+    // finalizers attached.
+    // Resources are filtered to be owned by the controller.
+    DESTROY_READY = 2;
```
andrewrynhard, line 139: This feels a little out of place here, but I am not comfortable enough with the code yet to suggest anything different.

smira: I was going back and forth on the design here, as it certainly feels like an output property, but outputs don't have namespaces, so I got stuck as a watch implies having a namespace.

smira: probably we can come up with some other concept here to express it better

api/v1alpha1/runtime.proto
```diff
@@ -127,6 +127,16 @@ enum ControllerInputKind {
     //
     // Controller can add finalizers on strong inputs.
     STRONG = 1;
+    // DestroyReady inputs provide notifications to the controllers when some resource
+    // is ready to be destroyed (all finalizers are removed).
+
+    // Usually outputs are declared as DestroyReady inputs to trigger reconcile loop
+    // when resource is ready to be destroy.
```
Unix4ever, line 134:
```suggestion
    // when resource is ready to be destroyed.
```

Unix4ever (approved):

api/v1alpha1/runtime.proto
andrewrynhard: Where did we land with this?

smira: I don't have any better idea than what we have here, it's more about the API, less about the actual implementation.

api/v1alpha1/runtime.proto
smira: thanks, fixed!

Unix4ever (approved):

PR #40, opened 2021-06-21
chore: re-run kres

AlekSi:
Updates Go version, etc.


smira (approved):

hack/release.toml
```diff
@@ -0,0 +1,16 @@
...
+commit = "HEAD"
+
+project_name = "runtime"
+github_repo = "cosi-project/runtime"
+match_deps = "^github.com/(cosi-project/[a-zA-Z0-9-]+)$"
+
+# previous = -
+# pre_release = true
+
+# [notes]
```
Unix4ever, line 16: I wonder if we should add some notes here. But if we're not going to release it, then it doesn't really matter.

Unix4ever (approved):

PR #41, opened 2021-06-21
feat: add sensitivity to RDs

AlekSi:
Sensitivity indicates how secret resource of this type is.
The empty value represents a non-sensitive resource.

### DIFF
--- a/pkg/resource/meta/resource_definition.go
+++ b/pkg/resource/meta/resource_definition.go
@@ -31,12 +31,16 @@ type PrintColumn struct {
 }
 
 // ResourceDefinitionSpec provides ResourceDefinition definition.
-type ResourceDefinitionSpec struct {
+type ResourceDefinitionSpec struct { //nolint:govet
 	Type             resource.Type      `yaml:"type"`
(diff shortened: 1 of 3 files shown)


pkg/resource/meta/resource_definition.go
```diff
@@ -31,12 +31,16 @@ type PrintColumn struct {
...
-type ResourceDefinitionSpec struct {
+type ResourceDefinitionSpec struct { //nolint:govet
 	Type             resource.Type      `yaml:"type"`
 	DisplayType      string             `yaml:"displayType"`
 	DefaultNamespace resource.Namespace `yaml:"defaultNamespace"`
 	Aliases          []resource.Type    `yaml:"aliases"`
 	PrintColumns     []PrintColumn      `yaml:"printColumns"`
+
+	// Sensitivity indicates how secret resource of this type is.
+	// The empty value represents a non-sensitive resource.
```
AlekSi, line 42: The alternative is to make the zero value mean "unspecified," but I would rather not introduce undefined behavior.

Unix4ever (approved):

smira (approved):

PR #51, opened 2021-07-06
fix: make default resource watch history more shallow

smira:
This reduce history depth from 1000 entries to 100 entries (for each
resource type under a namespace). The main benefit is memory savings.

Also gc unused and never implemented option.


Unix4ever (approved):

andrewrynhard (approved):
Something we can eventually make configurable?

smira:
> Something we can eventually make configurable?

Yep, that's a good idea, and I wonder if I should actually make it configurable in this PR

andrewrynhard (approved):

PR #69, opened 2022-04-12
refactor: remove fmt.Stringer from resource.Resource interface

smira:
It is not being used almost anywhere, and implementing it for every
resource is a lot of routine work.

It is now implemented in a generic way as a `resource.String` function.

### DIFF
--- a/pkg/controller/conformance/resources.go
+++ b/pkg/controller/conformance/resources.go
@@ -6,7 +6,6 @@ package conformance
 
 import (
 	"encoding/binary"
-	"fmt"
 
 	"github.com/cosi-project/runtime/pkg/resource"
 )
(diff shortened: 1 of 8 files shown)


pkg/resource/resource.go
```diff
@@ -61,3 +58,8 @@ func MarshalYAML(r Resource) (interface{}, error) {
 		Spec:     r.Spec(),
 	}, nil
 }
+
+// String returns representation suitable for %s formatting.
+func String(r Resource) string {
+	return fmt.Sprintf("%s(%s/%s)", r.Metadata().Type(), r.Metadata().Namespace(), r.Metadata().ID())
```
DmitriyMV, line 64: Maybe we can put Metadata in variable?
Like
```suggestion
	md := r.Metadata()
	return fmt.Sprintf("%s(%s/%s)", md.Type(), md.Namespace(), md.ID())
```

pkg/controller/conformance/resources.go
```diff
@@ -75,10 +74,6 @@ func (r *IntResource) SetValue(v int) {
 	r.value.value = v
 }
 
-func (r *IntResource) String() string {
```
DmitriyMV, line 78: Does new implementation print enough info? I see `r.value.value` here - is it important?

smira: it was more of a test thing I guess (for debugging), used only in the tests

DmitriyMV (approved):

PR #76, opened 2022-04-18
chore: migrate from AlekSi/pointer to siderolabs/go-pointer

DmitriyMV:
Now that we have generics and require Go 1.18 - redo pointer.* functions as two generic functions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,11 +7,11 @@ go 1.18
 replace gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20210315173758-8fb30b8e5a5b
 
 require (
-	github.com/AlekSi/pointer v1.1.0
 	github.com/cenkalti/backoff/v4 v4.1.1
 	github.com/gertd/go-pluralize v0.1.7
 	github.com/golang/protobuf v1.5.2
 	github.com/hashicorp/go-memdb v1.3.2
+	github.com/siderolabs/go-pointer v1.0.0
 	github.com/stretchr/testify v1.7.0
 	github.com/talos-systems/go-retry v0.3.1
 	go.uber.org/goleak v1.1.10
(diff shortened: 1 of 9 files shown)


pkg/ptr/ptr.go
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package ptr
+
+// To returns a pointer to a provided value
```
smira: nit:

```suggestion
// To returns a pointer to the provided value.
```

pkg/ptr/ptr.go
```diff
@@ -0,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package ptr
+
+// To returns a pointer to a provided value.
```
smira, line 7:
```suggestion
// To returns a pointer to the provided value.
```

pkg/ptr/ptr.go
```diff
@@ -0,0 +1,21 @@
...
+func To[T any](t T) *T {
+	return &t
+}
+
+// SafeDeref returns a pointer value or zelo value for a nil pointer.
```
smira, line 12:
```suggestion
// SafeDeref returns a pointer value or zero value for a nil pointer.
```

smira (commented):
should we use https://github.com/AlekSi/pointer/blob/main/generic.go#L7-L36 ?

pkg/ptr/ptr.go
DmitriyMV: Thx!

pkg/ptr/ptr.go
DmitriyMV: Thx!

pkg/ptr/ptr.go
DmitriyMV: Thx!

DmitriyMV:
> should we use https://github.com/AlekSi/pointer/blob/main/generic.go#L7-L36 ?
We could, but I don't see any reason too - since we are only using two small functions, I think it's better to have them in our own modules.

smira (approved):

PR #97, opened 2022-05-18
chore: provide an interface for getting wrapped protobuf value

Unix4ever:
This is required for Theila internal logic as it relies on dumping
protobuf specs as `map[string]interface{}` and sending it to the
javascript side.

With the change it will be possible to extract value from the wrapper
without knowing an exact type.

Signed-off-by: Artem Chernyshev <artem.chernyshev@talos-systems.com>

### DIFF
--- a/pkg/resource/protobuf/spec.go
+++ b/pkg/resource/protobuf/spec.go
@@ -12,6 +12,13 @@ type Spec[T any] interface {
 	*T
 }
 
+// ResourceSpecWrapper defines interface for the protobuf resource wrapped by a generic structure.
+type ResourceSpecWrapper interface {
+	ProtoMarshaler
+	ProtoUnmarshaler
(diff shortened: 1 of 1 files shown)


pkg/resource/protobuf/spec.go
```diff
@@ -40,6 +47,11 @@ func (spec *ResourceSpec[T, S]) UnmarshalProto(protoBytes []byte) error {
 	return proto.Unmarshal(protoBytes, spec.Value)
 }
 
+// Value returns wrapped protobuf object.
+func (spec *ResourceSpec[T, S]) Value() interface{} {
```
smira, line 51: should it be `proto.Message`?

Unix4ever: yep, sure

smira (approved):

PR #105, opened 2022-06-02
feat: implement resource metadata labels

smira:
Labels are simple key-value pairs attached to the resource metadata.
Metadata value can be still copied, labels "fork" themselves on update.

List request include new ListOption to specify label query: query terms
are combined with AND operator, and each term might check for label
existence or value match.


Unix4ever (approved):
:cool:

pkg/resource/labels.go
```diff
@@ -0,0 +1,109 @@
...
+// Raw returns the labels map.
+//
+// Label map should not be modified outside of the call.
+func (labels *Labels) Raw() map[string]string {
+	return labels.m
+}
+
+// Equal checks label for equality.
+func (labels Labels) Equal(other Labels) bool {
+	if labels.m == nil && other.m == nil {
```
DmitriyMV, line 61: Do we differentiate between nil map and empty map?

smira: that's a good question; also there's no way to delete labels right now, so there's no empty map ever.

smira: but I think we should handle that

smira: added the handling and a delete method

PR #109, opened 2022-06-06
feat: support listing resources which do not have a specific label

Unix4ever:
Introduce another query op: `NOT_EXISTS`.
In this mode the list will only return the resources which do not have
the specified label.

Signed-off-by: Artem Chernyshev <artem.chernyshev@talos-systems.com>


pkg/controller/protobuf/server/server.go
```diff
@@ -295,6 +295,8 @@ func (runtime *Runtime) List(req *v1alpha1.RuntimeListRequest, srv v1alpha1.Cont
 					opts = append(opts, state.WithLabelEqual(term.Key, term.Value))
 				case v1alpha1.LabelTerm_EXISTS:
 					opts = append(opts, state.WithLabelExists(term.Key))
+				case v1alpha1.LabelTerm_NOT_EXISTS:
```
smira, line 298: for both protobuf controller and state client, there is an opposite translation from COSI to protobuf

Unix4ever: yeah, linter got me for that :) Fixing

smira (approved):

frezbo (approved):

PR #110, opened 2022-06-07
feat: implement Watch with label selectors

smira:
COSI state can now watch list of resources with filters (based on label
selectors): events are filtered based on the label selectors, and if
change to the labels makes resource match/stop matching the query,
synthetic create/destroy event is pushed.

As part of the change, COSI now also publishes old resource state in
update events.


pkg/state/options.go
```diff
@@ -183,3 +184,41 @@ func WithKindTailEvents(n int) WatchKindOption {
 		opts.TailEvents = n
 	}
 }
+
+// WatchWithLabelQuery appends a label query to the list options.
+func WatchWithLabelQuery(query resource.LabelQuery) WatchKindOption {
```
Unix4ever, line 189: I wish we could reuse the same methods for list and watch as now the API will look a bit unobvious.
There's `WithLabelQuery` which only works with the list requests. 
Or we should rename `WithLabelQuery` to `ListWithLabelQuery`.

Or another idea is to have subfolder `watch` to make it be `watch.WithLabelQuery`.

pkg/state/protobuf/client/client.go
```diff
@@ -384,6 +375,20 @@ func watchAdapter(ctx context.Context, cli v1alpha1.State_WatchClient, ch chan<-
 			return
 		}
 
+		if msg.Event.Old != nil {
+			unmarshaled, err = protobuf.Unmarshal(msg.Event.Old)
+			if err != nil {
+				// no way to signal error here?
```
Unix4ever, line 381: maybe COSI resource should also have the error field in the metadata ;)

pkg/state/options.go
smira: it's still possible to do that bypassing `With` methods, I will think about better option tomorrow... I don't like it either

Unix4ever (approved):

PR #112, opened 2022-06-22
 feat: add type safe wrappers for state.Watch* methods

sauterp:
The introduction of generics makes it possible to have watch methods
that do type assertions automatically before returning resources to the
user. We add a generic wrapper for state.{Watch, WatchFor, WatchKind}.
In case a tombstone is encountered the zero value of the expected type
is returned.

### DIFF
--- a/pkg/safe/state.go
+++ b/pkg/safe/state.go
@@ -12,9 +12,11 @@ import (
 	"github.com/cosi-project/runtime/pkg/state"
 )
 
-// StateGet is a type safe wrapper around state.Get.
-func StateGet[T resource.Resource](ctx context.Context, st state.State, ptr resource.Pointer, options ...state.GetOption) (T, error) { //nolint:ireturn
-	got, err := st.Get(ctx, ptr, options...)
+func typeMismatchErr(expected, got any) error {
+	return fmt.Errorf("type mismatch: expected %T, got %T", expected, got)
+}
+
+func typeAssertOrZero[T resource.Resource](got resource.Resource, err error) (T, error) { //nolint:ireturn
 	if err != nil {
 		var zero T
 
@@ -25,10 +27,17 @@ func StateGet[T resource.Resource](ctx context.Context, st state.State, ptr reso
 	if !ok {
 		var zero T
 
-		return zero, fmt.Errorf("type mismatch: expected %T, got %T", result, got)
+		return zero, typeMismatchErr(result, got)
 	}
 
 	return result, nil
+}
+
+// StateGet is a type safe wrapper around state.Get.
+func StateGet[T resource.Resource](ctx context.Context, st state.State, ptr resource.Pointer, options ...state.GetOption) (T, error) { //nolint:ireturn
+	got, err := st.Get(ctx, ptr, options...)
+
+	return typeAssertOrZero[T](got, err)
 }
 
 // StateGetResource is a type safe wrapper around state.Get which accepts typed resource.Resource and gets the metadata from it.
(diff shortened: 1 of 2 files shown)


pkg/safe/state.go
```diff
@@ -87,6 +84,72 @@ func StateList[T resource.Resource](ctx context.Context, st state.State, ptr res
...
+			if event.Old != nil {
+				result.Old, ok = event.Old.(T)
+				if !ok {
+					fmt.Println(typeMismatchErr(result.Old, event.Old))
+				}
+			}
+
+			result.Type = event.Type
+
+			eventCh <- result
```
smira, line 113: this should be protected with `select` on `ctx.Done()`

pkg/state/state.go
```diff
@@ -28,13 +28,16 @@ func (eventType EventType) String() string {
 	return [...]string{"Created", "Updated", "Destroyed"}[eventType]
 }
 
-// Event is emitted when resource changes.
-type Event struct {
-	Resource resource.Resource
-	Old      resource.Resource
+// GenericEvent is emitted when resource changes.
+type GenericEvent[T resource.Resource] struct {
```
smira, line 32: `TypedEvent` instead of `GenericEvent` ?

pkg/safe/state.go
```diff
@@ -87,6 +84,79 @@ func StateList[T resource.Resource](ctx context.Context, st state.State, ptr res
...
+			}
+
+			var result state.TypedEvent[T]
+
+			ok := false
+
+			if event.Resource != nil {
+				result.Resource, ok = event.Resource.(T)
+				if !ok {
+					fmt.Println(typeMismatchErr(result.Resource, event.Resource))
```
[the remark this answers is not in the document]
smira: at least doing `fmt.Println()` is certainly not a way to go... the Watch interface doesn't support returning errors, so we either do getters like Dmitry suggested, or we put an error to the event itself?

pkg/safe/state.go
```diff
@@ -87,6 +84,86 @@ func StateList[T resource.Resource](ctx context.Context, st state.State, ptr res
...
 
+// WrappedStateEvent holds a state.Event that can be cast to its original Resource type when accessed with Event().
+type WrappedStateEvent[R resource.Resource] struct {
+	event state.Event
+}
+
+func (wse *WrappedStateEvent[T]) Event() (*state.TypedEvent[T], error) {
+	var result state.TypedEvent[T]
+
+	ok := false
```
smira, line 95:
```suggestion
	var ok bool
```

pkg/safe/state.go
```diff
@@ -87,6 +84,86 @@ func StateList[T resource.Resource](ctx context.Context, st state.State, ptr res
 	return NewList[T](got), nil
 }
 
+// WrappedStateEvent holds a state.Event that can be cast to its original Resource type when accessed with Event().
+type WrappedStateEvent[R resource.Resource] struct {
+	event state.Event
+}
+
+func (wse *WrappedStateEvent[T]) Event() (*state.TypedEvent[T], error) {
```
smira, line 92: probably it makes more sense to return by value here?

pkg/safe/state.go
```diff
@@ -87,6 +84,86 @@ func StateList[T resource.Resource](ctx context.Context, st state.State, ptr res
...
+	return &result, nil
+}
+
+func watch[T resource.Resource](ctx context.Context, eventCh chan<- WrappedStateEvent[T], untypedEventCh <-chan state.Event) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case event := <-untypedEventCh:
+			if resource.IsTombstone(event.Resource) {
```
smira, line 120: another option is to add getters for `Old()` and `Resource()` to the WrappedStateEvent, and this way we can pass through tombstones, they just won't be convertible. 

this way we don't need `TypedEvent` at all?

smira (approved):
lgtm!

PR #134, opened 2022-08-25
feat: prepare for etcd backend

utkuozdemir:
This PR contains the groundwork needed from the COSI runtime to support etcd as an additional backend:
- Remove `BumpVersion()` method on the resource metadata, since versions will now be bumped by the backend implementations and not on the consumer side
- Add a `SetUpdated(...)` method to metadata to be able to modify the modification date externally
- Add helper methods to `Version` to be able to easily get and manipulate versions
- Add a new error `ErrUnsupported` for operations that are not supported by a backend
- Update inmem implementation to increment resource versions on `Create` and `Update` similarly to etcd versioning
- Update conformance tests to be compatible with these changes

Signed-off-by: Utku Ozdemir <utku.ozdemir@siderolabs.com>


pkg/resource/version.go
```diff
@@ -26,6 +26,18 @@ var (
 
 const undefinedVersion = "undefined"
 
+// Value returns the underlying version number. Returns 0 if the version is undefined.
+func (v Version) Value() uint64 {
+	return pointer.SafeDeref(v.uint64)
+}
+
+// Next returns a new incremented version.
+func (v Version) Next() Version {
```
frezbo, line 35: is it expected to update version when the existing is `zero` i.e: undefined :thinking:

pkg/state/conformance/state.go
```diff
@@ -680,7 +685,12 @@ func (suite *StateSuite) TestWatch() {
 	// get event history
 	chWithTail := make(chan state.Event)
 
-	suite.Require().NoError(suite.State.Watch(ctx, path1.Metadata(), chWithTail, state.WithTailEvents(1000)))
+	err := suite.State.Watch(ctx, path1.Metadata(), chWithTail, state.WithTailEvents(1000))
+	if state.IsUnsupportedError(err) {
+		return
```
frezbo, line 690: so I guess we don't run other tests if the backend doesn;t support watching?

utkuozdemir: Watch is supported, only `tailEvents` (history replay) isn't. It is easier to review this hand in hand with the other PR: https://github.com/cosi-project/state-etcd/pull/1/files#diff-f4f4e768803041fd484adb327aebc9e2d23af18dbf33b854a0489207032c3940R281-R283

utkuozdemir:
See the actual etcd backend PR as well: https://github.com/cosi-project/state-etcd/pull/1

pkg/state/conformance/state.go
```diff
@@ -236,18 +235,15 @@ func (suite *StateSuite) TestWatchKind() {
 		suite.FailNow("timed out waiting for event")
 	}
 
-	oldVersion := path2.Metadata().Version()
-	path2.Metadata().BumpVersion()
-
-	suite.Require().NoError(suite.State.Update(ctx, oldVersion, path2))
+	suite.Require().NoError(suite.State.Update(ctx, resource.NewVersion(1), path2))
```
smira, line 238: we should be using `oldVersion` here still?

pkg/state/conformance/state.go
```diff
@@ -236,18 +235,15 @@ func (suite *StateSuite) TestWatchKind() {
...
 
 	select {
 	case event := <-ch:
 		suite.Assert().Equal(state.Updated, event.Type)
 		suite.Assert().Equal(resource.String(path2), resource.String(event.Resource))
-		suite.Assert().Equal(path2.Metadata().Version(), event.Resource.Metadata().Version())
+		suite.Assert().Equal(resource.NewVersion(2), event.Resource.Metadata().Version())
```
smira, line 244: same question here - why explicit version?

pkg/state/conformance/state.go
```diff
@@ -292,15 +285,20 @@ func (suite *StateSuite) TestWatchKind() {
...
+		suite.Assert().Equal(resource.NewVersion(3), event.Resource.Metadata().Version())
 	case <-time.After(time.Second):
 		suite.FailNow("timed out waiting for event")
 	}
 
 	// get event history
 	chWithTail := make(chan state.Event)
 
-	suite.Require().NoError(suite.State.WatchKind(ctx, path1.Metadata(), chWithTail, state.WithKindTailEvents(1000)))
+	err = suite.State.WatchKind(ctx, path1.Metadata(), chWithTail, state.WithKindTailEvents(1000))
```
smira, line 296: probably it makes more sense to split the test and `t.Skip()` it if not supported? it's more explicit and visible?

pkg/state/conformance/state.go
```diff
@@ -548,11 +546,17 @@ func (suite *StateSuite) TestWatchFor() {
 	wg.Wait()
 
 	suite.Require().NoError(err)
-	suite.Assert().Equal(r.Metadata().String(), path1.Metadata().String())
+
+	suite.Assert().Equal(r.Metadata().Type(), path1.Metadata().Type())
```
smira, line 550: not sure why this has to be changed now?

pkg/state/impl/inmem/collection.go
```diff
@@ -106,27 +107,29 @@ func (collection *ResourceCollection) inject(resource resource.Resource) {
...
 	collection.mu.Lock()
 	defer collection.mu.Unlock()
 
-	if _, exists := collection.storage[resource.Metadata().ID()]; exists {
-		return ErrAlreadyExists(resource.Metadata())
+	if _, exists := collection.storage[res.Metadata().ID()]; exists {
+		return ErrAlreadyExists(res.Metadata())
 	}
 
+	res.Metadata().SetVersion(resource.NewVersion(1))
```
smira, line 124: we do `DeepCopy` as we store, but we should set the version before we do `DeepCopy`, or rather `DeepCopy` after we set the version, so that he `res` upon return from `Create` should contain correct `Metadata()`

pkg/state/impl/inmem/collection.go
```diff
@@ -160,6 +159,9 @@ func (collection *ResourceCollection) Update(ctx context.Context, curVersion res
 		return ErrPhaseConflict(curResource.Metadata(), *options.ExpectedPhase)
 	}
 
+	newResource.Metadata().SetVersion(curVersion.Next())
```
smira, line 162: same as `create` here - we should also update the returned resource

pkg/state/conformance/state.go
```diff
@@ -847,5 +847,15 @@ func (suite *StateSuite) TestLabels() {
 	suite.Require().NoError(err)
 
 	suite.Require().Equal(1, list.Len())
-	suite.Assert().True(resource.Equal(path3, list.Get(0)))
```
smira, line 850: this is same issue - returned objects should contain correct version, no need to do any workarounds

smira:
tl;dr: `Update` and `Create` should return a resource with updated `Metadata().Version()`

pkg/state/conformance/state.go
utkuozdemir: It's to compare without the version - type, ns, id but not version

smira: version should match as well!

pkg/controller/protobuf/client/client.go
```diff
@@ -546,8 +546,7 @@ func (ctrlAdapter *controllerAdapter) Update(ctx context.Context, curVersion res
 	_, err = ctrlAdapter.adapter.client.Update(ctx, &v1alpha1.RuntimeUpdateRequest{
```
smira, line 546: to make things consistent, we should return new resource from the `Update` gRPC back?

pkg/state/conformance/state.go
utkuozdemir: done

pkg/resource/version.go
utkuozdemir: As we discussed, `zero` means non-existent and undefined means "I don't know about its version". Calling `.Next()` here will create a new version object for you, which is incremented by one on the called version. If called version is undefined, it will be treated as zero here.

pkg/controller/protobuf/client/client.go
utkuozdemir: done

pkg/state/impl/inmem/collection.go
```diff
@@ -106,51 +107,66 @@ func (collection *ResourceCollection) inject(resource resource.Resource) {
...
 
+	version, err := resource.ParseVersion("1")
+	if err != nil {
+		return err
+	}
+
+	updated := time.Now()
+
+	resCopy.Metadata().SetVersion(version)
+	resCopy.Metadata().SetUpdated(updated)
```
smira, line 132: we can probably skip setting `Updated` here, as `NewMetadata` automatically sets `created = updated = now()`

smira (approved):
:+1:

frezbo (approved):

pkg/state/impl/inmem/collection.go
utkuozdemir: done

PR #144, opened 2022-09-06
chore: move "implements" checks to compile time

DmitriyMV:
There is no need to use `assert.Implements` since we can express this check during compile time. Go will eliminate `_` variables and any accompanying allocations during dead-code elimination phase.


smira:
actually we moved to `assert.Implements` from using `_` form assertions; there was a reason, and I don't remember now the exact one

pkg/controller/runtime_test.go
```diff
@@ -5,17 +5,11 @@
...
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/cosi-project/runtime/pkg/state/impl/inmem"
 )
 
-func TestInterface(t *testing.T) {
-	t.Parallel()
-
-	assert.Implements(t, (*controller.Reader)(nil), state.WrapCore(&inmem.State{}))
-}
+// We can move this check to [inmem.State] but it will result in allocation, since Go cannot currenlty
```
frezbo, line 13:
```suggestion
// We can move this check to [inmem.State] but it will result in allocation, since Go cannot currently
```

DmitriyMV: Thanks, fixed!

smira:
https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables

DmitriyMV:
@smira 
> actually we moved to assert.Implements from using _ form assertions; there was a reason, and I don't remember now the exact one

The only thing I can think of is that is can result in cyclic dependencies when implementation defines things which interface package uses. Another thing, which is mostly cosmetic, is that implementation becomes aware of interface package. But judging from the diff, no new imports were added, so we should be good?

DmitriyMV:
@smira 
> https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables

I don't agree on this with Dave. And even he says that it's a "thought experiment". The "blank-variable-assignment" technique is pretty much used everywhere in Go code - many generators use them. Mat Ryer - author of Go Time podcast - [supports them](https://medium.com/@matryer/golang-tip-compile-time-checks-to-ensure-your-type-satisfies-an-interface-c167afed3aae). [Josh Bleecher Snyder](https://commaok.xyz/post/safer-generated-code/) who is contributor to Go compiler is also supporting checks like this.

At the of the day I believe that while we lack any "const-expr" checks, moving checks from compile time to runtime is never a bad thing. In future versions of Go if/when we get compile time magic checks we might reconsider.

DmitriyMV:
There is also [this](https://github.com/golang/go/issues/34868) discussion in Go issue tracker about making init functions run in compile time in certain scenarios.

smira:
I just don't see a benefit in going back and forth without any clear value. Next year we might flip it backwards. 

What is the problem we're solving with this change? Is there any real benefit?

smira (approved):
discussed in a call

PR #148, opened 2022-09-13
fix: ensure there is no deadlock in ResourceCollection.Watch and WatchAll

DmitriyMV:
Also remove conformance.StateSuite.TearDownTest method.


smira (approved):
:cool:

PR #149, opened 2022-09-15
feat: add encryption.Marshaler

DmitriyMV:
This `Marshaler` is used to encrypt data that was marshaled by the underlying `store.Marshaler`.

### DIFF
--- a/pkg/state/impl/store/encryption/marshaler.go
+++ b/pkg/state/impl/store/encryption/marshaler.go
@@ -1,0 +1,176 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package encryption provides encryption support for [store.Marshaler].
+package encryption
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"fmt"
+	"io"
+	"sync"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state/impl/store"
+)
+
+// Marshaler encrypts and decrypts data from the underlying marshaler.
+type Marshaler struct {
+	underlying store.Marshaler
+	cipher     *Cipher
+}
+
+// NewMarshaler creates new Marshaler.
+func NewMarshaler(m store.Marshaler, c *Cipher) *Marshaler {
+	return &Marshaler{underlying: m, cipher: c}
+}
+
+// MarshalResource implements Marshaler interface.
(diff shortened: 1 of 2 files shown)


DmitriyMV:
I also intend to add tests similar to `TestBboltConformance`

pkg/state/impl/store/encryption/marshaler.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to init cipher: %w", err)
+	}
+
+	nonce := make([]byte, 12)
+	if _, err := io.ReadFull(rand.Reader, nonce); err != nil {
+		return nil, fmt.Errorf("failed to generate nonce: %w", err)
+	}
+
+	// We attach noonce before the encrypted data
```
smira, line 115:
```suggestion
	// We attach nonce before the encrypted data
```

pkg/state/impl/store/encryption/marshaler_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/state/conformance"
+	"github.com/cosi-project/runtime/pkg/state/impl/store"
+	"github.com/cosi-project/runtime/pkg/state/impl/store/encryption"
+)
+
+func TestMain(m *testing.M) {
+	err := protobuf.RegisterResource(conformance.PathResourceType, &conformance.PathResource{})
```
smira, line 22: can't we do this via a more regular `init` function?

pkg/state/impl/store/encryption/marshaler.go
```diff
@@ -0,0 +1,169 @@
...
+	encrypted := aead.Seal(nonce, nonce, b, nil)
```
smira, line 116: should we also reserve something for the format/type extension?

e.g. we could do a single `\0` byte at the beginning, and expand this in the future to handle key IDs or different encryption formats, etc.

DmitriyMV: Yeah, I think its a good idea.

DmitriyMV: Thanks, fixed!

pkg/state/impl/store/encryption/marshaler_test.go
DmitriyMV: Hmm, there can be several `init`s, while there can be only one `TestMain` per package. But I agree - its overcomplication in this case.

pkg/state/impl/store/encryption/marshaler.go
DmitriyMV: Fixed!

smira (approved):

PR #153, opened 2022-09-19
feat: implement resource metadata annotations

### LINKED ISSUE #135: add 'annotations' to resource metadata
Same as 'labels', but no way to filter on annotations.

smira:
The implementation is exactly same for labels and annotations, but annotations don't support any kind of querying/filtering.

Fixes #135


smira:
See also: https://github.com/cosi-project/specification/pull/4

pkg/resource/internal/kv/kv.go
```diff
@@ -0,0 +1,161 @@
...
+func (kv *KV) Delete(key string) {
+	if _, ok := kv.m[key]; !ok {
+		// no change
+		return
+	}
+
+	kvCopy := make(map[string]string, len(kv.m))
+
+	for k, v := range kv.m {
+		kvCopy[k] = v
```
frezbo, line 35: is it easier to check if the key to delete matches here and just continue?

frezbo (approved):

pkg/resource/internal/kv/kv.go
smira: fixed!

frezbo (approved):

Issue #131, opened 2022-08-15
encrypting marshaled resource data


smira:
The high-level overview is to provide an option to encrypt resource data as it gets marshaled to bytes for subsequent storage.

The potential integration point is around those two methods:

https://github.com/cosi-project/runtime/blob/b4c8d5f13338031ca64d2bbeac044e8d94681d61/pkg/state/impl/store/store.go#L10-L15

There should be a master encryption key which is used to encrypt the data. We should allow storing encryption metadata within the encrypted data, so that we can change the encryption format in the future, e.g.:

```
<encryption-format> | <key-id> | <encrypted data...>
```

The master encryption key should be unlocked indirectly via the encryption key slots (similar to LUKS format), so that we can rotate the keys used to unlock the master key without ever changing the encryption key.

```
<key-slot-1> | <key-slot-2> | ...
```

Each key slot contains the master key encrypted with some encryption key. Any encryption key valid for any slot allows to unlock the master encryption key and decrypt the storage.

Key slots should support the validity check - if the encryption key doesn't match, there should be an explicit error vs. trying to use corrupted master key.

### Initializing encryption

COSI generates random master key.

COSI is given one or more encryption keys, and these are used to encrypt the master key to generate the key slot data.

COSI starts encrypting resource data using the master key (which is only stored in memory)

### Unlocking the storage

COSI is given the encryption key and slot ID. COSI decrypts the master key using the encryption key.

COSI starts encrypting/decrypting the data using the master key.

### Slot operations

Once the master key is unlocked, it should be possible to alter the encryption key slots - e.g. to add/remove slots, change the encryption key. For the additional protection encryption key should be validated before slot operation succeeds. Same with adding a new slot - an existing encryption key should be given for other slot to allow adding a new slot.

DmitriyMV:
Encryption part is done in https://github.com/cosi-project/runtime/pull/149.

PR #156, opened 2022-09-22
feat: implement key storage

DmitriyMV:
This commits adds a new type - keystorage.KeyStorage which is used to safely store master key.

For #131


pkg/keystorage/keystorage.go
```diff
@@ -0,0 +1,193 @@
...
+// KeyStorage is a key storage that can be used to store and retrieve the master key.
+//
+//nolint:govet
+type KeyStorage struct {
+	mx         sync.Mutex
+	underlying key_storage.Storage
+}
+
+// SetMasterKey sets the master key for the key storage, encrypts it using public key and stores it in slot id.
+func (ks *KeyStorage) SetMasterKey(masterKey []byte, slotID, slotPublicKey string) error {
```
smira, line 29: suggestion: we could call this `Initialize()` (or something similar) and make this function generate the random `masterKey` itself

pkg/keystorage/keystorage.go
```diff
@@ -0,0 +1,193 @@
...
+		EncryptedKey: []byte(encryptedSlot),
+	}
+
+	ks.underlying.KeysHmacHash = ks.hashSlots(masterKey)
+
+	return nil
+}
+
+// GetKey returns the attempts to decrypt the key slot with the given private key and return the master key.
+func (ks *KeyStorage) GetKey(slotID, slotPrivateKey string) ([]byte, error) {
```
smira, line 102: I think this returns the master key, to distinguish from `SetKey`, which operates on key slots

```suggestion
func (ks *KeyStorage) GetMasterKey(slotID, slotPrivateKey string) ([]byte, error) {
```

pkg/keystorage/keystorage.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	ks.underlying.KeysHmacHash = ks.hashSlots(masterKey)
+
+	return nil
+}
+
+// SetKey sets the new key slot with the given slot id and public key using previus slot and its private key.
+// It cannot be used to update an existing key slot.
+// It is required to call SetMasterKey() or UnmarshalBinary() to initialize the key storage first.
+func (ks *KeyStorage) SetKey(newSlotID, newSlotPublicKey, oldSlotID, oldSlotPrivateKey string) error {
```
smira, line 66: we probably need `DeleteSlot` function as well

and not sure about proper naming here, `EncryptSlot` maybe?

smira (approved):
my concern is that this doesn't include actual writes/reads to the database, so it might be hard to integrate this with the conflict check if it's done outside of this type -> e.g. if adding the key results in a conflict, the change should be reverted?

pkg/keystorage/keystorage.go
DmitriyMV: Renamed to `Initialize` and added a new method `InitializeRnd`.

pkg/keystorage/keystorage.go
DmitriyMV: Done!

pkg/keystorage/keystorage.go
DmitriyMV: Changed to `AddKeySlot` and added `DeleteKeySlot`

PR #158, opened 2022-09-29
feat: implement generic (template) transform controller

smira:
This type of controller transforms (or maps) resources of one type into resources of another type. Transformation function might have side effects which provide the additional feature.

Also additional inputs might be used to facilitate the transformation process.

Controller provides the necessary boilterplate to handle all flows with finalizers and teardown flow.

### DIFF
--- a/pkg/controller/generic/transform/controller.go
+++ b/pkg/controller/generic/transform/controller.go
@@ -1,0 +1,369 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package transform
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/siderolabs/gen/xerrors"
+	"go.uber.org/zap"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/controller/generic"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+)
+
+// Controller provides a generic implementation of a controller which implements controller transforming Input resources into Output resources.
+//
+// Controller supports full flow with finalizers:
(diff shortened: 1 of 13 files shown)


frezbo (approved):

pkg/resource/rtestutils/assertions.go
```diff
@@ -0,0 +1,134 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package rtestutils
```
DmitriyMV, line 5: Maybe something like `testresource`? Because right now some functions read like "resource test utils assert resource" which sounds a bit redundant.

pkg/controller/generic/transform/controller.go
```diff
@@ -0,0 +1,217 @@
...
+			Type: output.ResourceDefinition().Type,
+			Kind: controller.OutputExclusive,
+		},
+	}
+}
+
+// Run implements controller.Controller interface.
+//
+//nolint:gocognit,gocyclo,cyclop
+func (ctrl *Controller[Input, Output]) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
```
DmitriyMV, line 97: Run is incredibly hard to read. Maybe we should finally split into separate functions, since we are already refactoring it?

DmitriyMV (approved):
I think all is good except for Run method.

pkg/controller/generic/transform/controller.go
smira: yeah, I agree, next problem is to figure out _how_

pkg/resource/rtestutils/assertions.go
smira: I thought `testresource` is too generic, even more weird `testresource.AssertResources`.

`testutils` is nice, but there might be multiple of them, so prefixing with `r` to make it unique

DmitriyMV: `testresource.Assert` is good no?

smira: doesn't quite sound good to me, there might more similar functions, like getting a list of resource IDs. I moved only _some_ functions for now, let's see what will be needed

PR #187, opened 2022-11-25
feat: support new Bootstrapped and Errored watch events

### LINKED ISSUE #175: Watch API improvements
Introduce new types of Watch events:

* Error to pass through errors, we need to use it in the gRPC wrapper, in the etcd state implementation; inmem implementation can use it to report buffer overrun
* Bootstrapped to report when initial contents are sent (only when BootstrapContents is requested)

Figure out how to introduce it without breaking clients... New Watch call? add options to the existing one?

smira:
In the regular COSI, new event types are enabled immediately.

With protobuf client, new types are enabled when using new COSI client, old COSI client against new server won't receive new events, as old implementation panics on nil Resource in the event structure.

Fixes #175


smira:
See https://github.com/cosi-project/specification/pull/5

TimJones (approved):

pkg/state/impl/inmem/collection.go
```diff
@@ -455,10 +465,15 @@ func (collection *ResourceCollection) WatchAll(ctx context.Context, ch chan<- st
 			}
 
 			if collection.writePos-pos >= int64(collection.capacity) {
-				// buffer overrun, there's no way to signal error in this case,
-				// so for now just return
 				collection.mu.Unlock()
 
+				select {
+				case ch <- state.Event{
+					Type: state.Errored,
```
utkuozdemir, line 472: why don't we fill the error field when the state is errored?

smira: :facepalm:  thanks for noticing that, I will write a test for it!

utkuozdemir: I think the assertions need to be inverted as well :)

utkuozdemir (approved):
🆒

smira:
I addressed the feedback, and added tests for Watch errors (simulated buffer overrun), and also controller-runtime now fails if any internal watch call fails (also with a unit-test)

PR #193, opened 2022-12-08
chore: redo YAML `inline` as MarshalYAML/UnmarshalYAML instead of the field tag.

DmitriyMV:
`yaml:",inline"` do not play well with custom types `MarshalYAML`, so for `ResourceSpec` we have to implement it manually.

### DIFF
--- a/pkg/resource/protobuf/spec_test.go
+++ b/pkg/resource/protobuf/spec_test.go
@@ -5,10 +5,13 @@
 package protobuf_test
 
 import (
+	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
+	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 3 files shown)


pkg/resource/protobuf/spec_test.go
```diff
@@ -128,12 +131,108 @@ func (t ReflectSpec) DeepCopy() ReflectSpec {
...
+
+	if !resource.Equal(original, result) {
+		t.Log("original ->", must(yaml.Marshal(original.Spec()))(t))
+		t.Log("result ->", must(yaml.Marshal(result.Spec()))(t))
+		t.FailNow()
+	}
+}
+
+func ExampleResource_testInline() {
+	// This is a test to esnure that our custom 'inline' does work.
```
utkuozdemir, line 178:
```suggestion
	// This is a test to ensure that our custom 'inline' does work.
```

smira (approved):

PR #196, opened 2022-12-09
feat: implement dynamic sizing of history buffer for in-mem resources

smira:
It's now possible to set initial/max capacity of the event buffer for in-memory resource collection. The defaults are still fixed 100 entries.

Allow to optionally set dynamic capacity of the history buffer, so that for resource types which are not updated frequently less memory is wasted for the allocated event stream, while frequently updated resources might benefit from a deeper history buffer.

This should help Talos, as it has 87 different resource types (and at least each resource has its own buffer), while some resources are updated frequently, while others are almost static. We can set dynamic capacity with smaller initial capacity which will grow on demand based on the number resources created/updated.

### DIFF
--- a/pkg/state/impl/inmem/collection.go
+++ b/pkg/state/impl/inmem/collection.go
@@ -34,20 +34,22 @@ type ResourceCollection struct {
 
 	writePos int64
 
-	capacity int
-	gap      int
+	capacity    int
+	maxCapacity int
+	gap         int
 }
 
 // NewResourceCollection returns new ResourceCollection.
-func NewResourceCollection(ns resource.Namespace, typ resource.Type, capacity, gap int, store BackingStore) *ResourceCollection {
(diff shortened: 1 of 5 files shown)


frezbo (approved):

pkg/state/impl/inmem/collection.go
```diff
@@ -34,20 +34,22 @@ type ResourceCollection struct {
...
-		ns:       ns,
-		typ:      typ,
-		capacity: capacity,
-		gap:      gap,
-		storage:  make(map[resource.ID]resource.Resource),
-		stream:   make([]state.Event, capacity),
-		store:    store,
+		ns:          ns,
+		typ:         typ,
+		capacity:    initialCapacity,
```
DmitriyMV, line 47: I think in future we can replace this field by using `cap(stream)` directly.

DmitriyMV (approved):

PR #202, opened 2022-12-20
fix: always destroy tearing down outputs

smira:
The problem is triggered with the following flow: let's image transform controller maps resources A to resources B keeping ID the same.

1. A(1) is created.
2. Transform controller creates output B(1).
3. Something sets a finalizer on B(1).
4. A(1) is destroyed.
5. Transform controller can't destroy B(1), as it has a finalizer set, but it puts B(1) into tearing down phase.
6. A(1) is re-created (that is essentially new A(1)).
7. Transform controller can't update B(1), as it's being torn down.
8. Finalizer from B(1) is removed.
9. Transform controller still can't move on.

Before the fix controller will be "stuck" at the step (9), while with the fix controller will first destroy tearing down B(1), and then re-create it from new A(1).

### DIFF
--- a/pkg/controller/generic/transform/transform_test.go
+++ b/pkg/controller/generic/transform/transform_test.go
@@ -350,6 +350,65 @@ func TestDestroyReconcileTeardown(t *testing.T) {
 	})
 }
 
+func TestDestroyFinalizersRecreateInput(t *testing.T) {
+	setup(t, func(ctx context.Context, st state.State, runtime *runtime.Runtime) {
+		require.NoError(t, runtime.RegisterController(NewABController(nil)))
+
+		for _, a := range []*A{
+			NewA("1", ASpec{Int: 1}),
+			NewA("2", ASpec{Int: 1}),
+			NewA("3", ASpec{Int: 1}),
+		} {
+			require.NoError(t, st.Create(ctx, a))
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/generic/transform/transform_test.go
```diff
@@ -350,6 +350,65 @@ func TestDestroyReconcileTeardown(t *testing.T) {
...
+			},
+		)
+
+		// remove the finalizer
+		require.NoError(t, st.RemoveFinalizer(ctx, NewB("transformed-3", BSpec{}).Metadata(), finalizer))
+
+		// the output should be re-created with new spec
+		rtestutils.AssertResources(ctx, t, st, []resource.ID{"transformed-3"},
+			func(r *B, assert *assert.Assertions) {
+				assert.Equal(`""-2`, r.TypedSpec().Out)
```
frezbo, line 405: I didn't quite understand the asserting on `""-2` here and before, but nothing blocking.

frezbo (approved):

pkg/controller/generic/transform/transform_test.go
smira: we re-create A(3) resource with different Spec, so when transformed it should become `""-2` in the out spec. 

This way we check which generation of A(3) output B(transformed-3) belongs to (first it stick with initial input, and then moves on to the new output).

Unix4ever (approved):

PR #209, opened 2023-01-12
feat: add Set/Has methods to finalizers

smira:
Has was missing, and causing lots of trouble.

Set method allows to copy finalizers from one resource to another.

### DIFF
--- a/pkg/resource/finalizer.go
+++ b/pkg/resource/finalizer.go
@@ -46,3 +46,26 @@ func (fins *Finalizers) Remove(fin Finalizer) bool {
 func (fins Finalizers) Empty() bool {
 	return len(fins) == 0
 }
+
+// Has returns true if fin is present in the list of finalizers.
+func (fins Finalizers) Has(fin Finalizer) bool {
+	for _, f := range fins {
+		if f == fin {
+			return true
+		}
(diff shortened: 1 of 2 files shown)


pkg/resource/finalizer.go
```diff
@@ -46,3 +46,26 @@ func (fins *Finalizers) Remove(fin Finalizer) bool {
...
+		if f == fin {
+			return true
+		}
+	}
+
+	return false
+}
+
+// Set copies the finalizers from the other.
+func (fins *Finalizers) Set(other Finalizers) {
```
DmitriyMV, line 62: `CopyFrom` maybe? `Set` looks weird when there is `Has` which accepts only one `Finalizer`.

smira: as `Finalizers` is a kind of a set (but actually a slice), I think it is consistent. I think `Copy` in Go is more of a top-level function, rather than a function on a type.

DmitriyMV: Maybe it's a good idea to make `Copy(dst, src *Finalizers)`. But I don't think I have a strong opinion about this.

DmitriyMV (approved):

PR #213, opened 2023-01-18
fix: ignore other resources in rtestutils.Destroy

smira:
This shows up if there are multiple teardown/destroy sequences concurrently for the same resource type: `Destroy` might pick up a resource with the ID not from the list and destroy it (or return earlier than it should).

### DIFF
--- a/pkg/resource/rtestutils/destroy.go
+++ b/pkg/resource/rtestutils/destroy.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"testing"
 
+	"github.com/siderolabs/gen/slices"
 	"github.com/siderolabs/go-pointer"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


utkuozdemir (approved):

pkg/resource/rtestutils/destroy.go
```diff
@@ -37,15 +38,26 @@ func Destroy[R ResourceWithRD](ctx context.Context, t *testing.T, st state.State
...
+
+	for len(idMap) > 0 {
 		event, ok := recvWithContext(ctx, watchCh)
 		if !ok {
-			require.FailNow(t, "timeout", "left: %d %s", left, rds.Type)
+			require.FailNow(t, "timeout", "left: %d %s", len(idMap), rds.Type)
+		}
+
+		r, err := event.Resource()
+
```
DmitriyMV, line 50: Extra newline?

DmitriyMV (approved):

pkg/resource/rtestutils/destroy.go
smira: fair :)

frezbo (approved):

PR #219, opened 2023-02-03
feat: support query on resource IDs

smira:
Allow List/WatchKind results to be filtered on resource IDs.

See https://github.com/cosi-project/specification/pull/6


pkg/resource/id_query.go
```diff
@@ -0,0 +1,33 @@
...
+
+// IDQuery is the query on the resource ID.
+type IDQuery struct {
+	Regexp *regexp.Regexp
+}
+
+// Matches if the resource ID matches the ID query.
+func (query IDQuery) Matches(md Metadata) bool {
+	if query.Regexp == nil {
+		return true
```
frezbo, line 19: match true if it's nil? :thinking:

pkg/resource/metadata_test.go
```diff
@@ -187,7 +187,7 @@ finalizers:
 	assert.True(t, md.Equal(in))
 }
 
-var ts, _ = time.Parse(time.RFC3339, "2021-06-23T19:22:29Z")
+var ts, _ = time.Parse(time.RFC3339, "2021-06-23T19:22:29Z") //nolint:errcheck
```
DmitriyMV, line 190: Linter is probably right here - we do want to ensure that `err` is `nil` I suggest something like this
```suggestion
var ts = func() time.Time {
	res, err := time.Parse(time.RFC3339, "2021-06-23T19:22:29Z")
	if err != nil {
		panic(err)
	}

	return res
}()
```

frezbo (approved):

pkg/state/protobuf/server/server.go
```diff
@@ -75,6 +75,17 @@ func (server *State) List(req *v1alpha1.ListRequest, srv v1alpha1.State_ListServ
 				opts = append(opts, state.WithLabelQuery(labelOpts...))
 			}
 		}
+
+		if req.Options.GetIdQuery() != nil {
+			var idOpts []resource.IDQueryOption
```
DmitriyMV, line 80: Redundant declaration?

pkg/resource/id_query.go
smira: yeah, it means that no query was set

pkg/state/protobuf/server/server.go
```diff
@@ -75,6 +75,17 @@ func (server *State) List(req *v1alpha1.ListRequest, srv v1alpha1.State_ListServ
...
+
+			idOpts, err := ConvertIDQuery(req.Options.GetIdQuery())
```
DmitriyMV, line 82: This variant and the one below are different. Is there any reason for this?

pkg/resource/metadata_test.go
DmitriyMV: Okay, although I think if we can check things during assignments , we should do this. I usually just wrap the expression in `func must[T any](v T, err error) T`. Actually we should probably have `testutils` in `gen` so we could stop writing `must` for every package.

pkg/state/protobuf/server/server.go
smira: I guess I was refactoring it and missed it, not real difference, I will clean it up, thanks

Unix4ever (approved):

PR #237, opened 2023-03-09
feat: aggregate WatchKind and watch event deduplication

smira:
This supersedes #197 (combined into this PR).

The core problem solved in this PR is balancing the producers (updating resources) and consumers of watch events (controller runtime). With the existing code, producers can easily outperform producers and cause a buffer overrun.

We add new `WatchKindAggregate` method which retuns state events as a pack of several events which allows to receive more events in a single channel receive operation, giving leverage to consumers.

Under a burst of events related to small number of resources, deduper might aggregate several watch events into smaller number of final updates which trigger controller updates.


smira:
See https://github.com/cosi-project/specification/pull/7

pkg/controller/runtime/runtime.go
```diff
@@ -240,32 +246,136 @@ func (runtime *Runtime) watch(resourceNamespace resource.Namespace, resourceType
...
+	// deduplication process consists of two goroutines:
+	// 1. the first goroutine reads events from the watch channel as fast as possible,
+	//    reduces the event value to 'reducedMetadata' and pushes them to the map
+	// 2. the second goroutine consumes a map from the first goroutine, performs the work required
+	//    to trigger updates in the dependent controller
+	//
+	// The design idea is to consume watch events as fast as possible, while delaying "heavy" work
+	// to the second goroutine.
+	//
+	// There is a trick being used which sends a single map back and forth between the two goroutines.
```
andrewrynhard, line 266: Very cool

andrewrynhard (commented):
I took a look but out of the loop for some time so not a sharp eye. It gets my soft approval but I will wait for someone more familiar with the code to approve.

pkg/controller/runtime/runtime.go
```diff
@@ -240,32 +246,136 @@ func (runtime *Runtime) watch(resourceNamespace resource.Namespace, resourceType
...
+		var events []state.Event
+
+		// wait for an event
+		select {
+		case <-runtime.runCtx.Done():
+			return
+		case events = <-runtime.watchCh:
+		}
+
+		sourceEvents := 0
```
DmitriyMV, line 289: Unused variable - only writes and no reads. Is there any reason for it?

pkg/controller/runtime/runtime.go
```diff
@@ -240,32 +246,136 @@ func (runtime *Runtime) watch(resourceNamespace resource.Namespace, resourceType
...
+		case m = <-ch:
 		case <-runtime.runCtx.Done():
 			return
-		case e = <-runtime.watchCh:
 		}
 
-		if e.Type == state.Errored {
-			// watch failed, we need to abort
-			runtime.watchErrors <- e.Error
+		for _, e := range events {
```
DmitriyMV, line 301: I think we can use local function (closure?) here and remove code duplication.

pkg/controller/runtime/adapter.go
```diff
@@ -46,7 +46,25 @@ type adapter struct {
 	watchFilterMu sync.Mutex
 }
 
-type watchFilter func(*resource.Metadata) bool
+type reducedMetadata struct {
```
DmitriyMV, line 49: I think we can leave a note here, that `reducedMetadata` should always be comparable.

pkg/controller/runtime/runtime.go
smira: yeah, I was doing some debugging, and forgot to remove the leftovers

pkg/controller/runtime/adapter.go
smira: updated

pkg/controller/runtime/runtime.go
smira: fixed

pkg/state/impl/inmem/collection.go
```diff
@@ -440,23 +441,39 @@ func (collection *ResourceCollection) WatchAll(ctx context.Context, ch chan<- st
...
-			if !channel.SendWithContext(ctx, ch, state.Event{Type: state.Bootstrapped}) {
-				return
+				events = append(events, state.Event{Type: state.Bootstrapped})
+
+				if !channel.SendWithContext(ctx, aggCh, events) {
+					return
+				}
 			}
+
+			bootstrapList = nil
```
DmitriyMV, line 476: No more reads or writes after than, any reason for nilling it?

pkg/controller/runtime/runtime.go
```diff
@@ -240,32 +246,133 @@ func (runtime *Runtime) watch(resourceNamespace resource.Namespace, resourceType
...
 		controllers, err := runtime.depDB.GetDependentControllers(controller.Input{
-			Namespace: md.Namespace(),
-			Type:      md.Type(),
-			ID:        pointer.To(md.ID()),
+			Namespace: k.namespace,
+			Type:      k.typ,
+			ID:        pointer.To(k.id),
 		})
 		if err != nil {
 			// TODO: no way to handle it here
```
DmitriyMV, line 378: I think we can at least log it using `Runtime.logger`. This comment predates that field, so maybe we can change this too?

pkg/controller/runtime/runtime.go
```diff
@@ -240,32 +246,133 @@ func (runtime *Runtime) watch(resourceNamespace resource.Namespace, resourceType
...
 			return
 		}
 
-		md := e.Resource.Metadata()
+		// drain the watchCh by consuming all immediately available events
+	drainer:
+		for {
+			select {
+			case events = <-runtime.watchCh:
+				if !processEvents() {
```
DmitriyMV, line 323: I think we should pass `events` here explicitly. The reason is - this code can be easily broken if we change the line above to `:=`.  I would say the same applies to `m` except we get it only once, unlike `events`.

pkg/controller/runtime/runtime.go
```diff
@@ -240,32 +246,133 @@ func (runtime *Runtime) watch(resourceNamespace resource.Namespace, resourceType
...
+		}
+
+		// consume any first key of the map
+		var k reducedMetadata
+
+		for k = range m {
+			break
+		}
+
+		delete(m, k)
```
DmitriyMV, line 358: I think we can just add `takeOne` method to `dedup` type, since it is custom type anyway.

Something like:
```go
func (d dedup) takeOne() reducedMetadata {
	for k := range d {
		delete(d, k)
		return k
	}
}
```

```

DmitriyMV (approved):
Except few minor nitpicks, all good!

pkg/state/impl/inmem/collection.go
DmitriyMV: IIRC runtime/compiler knows that variable is no longer used if the function is big enough. Since `WatchAll` returns it means that only this goroutine could probably keep this variable.

Anyway - I don't mind keep that block, but maybe we should comment the reason for doing this?

pkg/controller/runtime/runtime.go
smira: fixed

pkg/controller/runtime/runtime.go
smira: good point, fixed

pkg/controller/runtime/runtime.go
smira: fixed

pkg/state/impl/inmem/collection.go
smira: yep, added

PR #243, opened 2023-03-15
feat: implement compressed resource marshaling

smira:
It is backwards compatible (i.e. it can read uncompressed resources), and it has a minimum size to be enabled.

This should help a lot with resources which have compressible data.


frezbo (approved):
I thought protobuf was already compressed format

andrewrynhard (approved):

PR #248, opened 2023-03-31
feat: add cleanup controller

DmitriyMV:
This PR adds new _generic_ cleanup controller. Currently, it is bundled with two implementations:
- One waits for outputs to be destroyed and only then allows the input to be destroyed.
- Another removes the outputs (ignoring ownership) itself when input is being removed.


pkg/controller/protobuf/client/client.go
```diff
@@ -609,7 +609,11 @@ func (ctrlAdapter *controllerAdapter) Modify(ctx context.Context, emptyResource
 	}
 }
 
-func (ctrlAdapter *controllerAdapter) Teardown(ctx context.Context, resourcePointer resource.Pointer) (bool, error) {
+func (ctrlAdapter *controllerAdapter) Teardown(ctx context.Context, resourcePointer resource.Pointer, opOpts ...controller.Option) (bool, error) {
+	if len(opOpts) > 0 {
+		return false, fmt.Errorf("options are not supported over the wire yet")
+	}
+
```
DmitriyMV, line 616: Do we need to fix this or Options can remain a special case for now?

pkg/controller/generic/cleanup/cleanup.go
```diff
@@ -0,0 +1,348 @@
...
+func (h *handler[I, O]) Outputs() []controller.Output {
+	if h.outputs != nil {
+		return h.outputs()
+	}
+
+	return nil
+}
+
+// HasNoOutputs is a helper function to create a [Handler] that waits for all outputs to be removed.
+func HasNoOutputs[O generic.ResourceWithRD, I generic.ResourceWithRD](
```
smira, line 214: how can we handle multiple outputs here? i.e. what if one resource depends on multiple output resources?

pkg/controller/runtime/adapter.go
```diff
@@ -338,11 +338,18 @@ func (adapter *adapter) Teardown(ctx context.Context, resourcePointer resource.P
 		return false, fmt.Errorf("resource %q/%q is not an output for controller %q, teardown attempted on %q", resourcePointer.Namespace(), resourcePointer.Type(), adapter.name, resourcePointer.ID())
 	}
 
-	return adapter.runtime.state.Teardown(ctx, resourcePointer, state.WithTeardownOwner(adapter.name))
+	var opts []state.TeardownOption
+
+	opOpt := controller.ToOptions(opOpts...)
+	if !opOpt.IgnoreOwner {
```
smira, line 344: this doesn't look like `IgnoreOwner`, but rather `WithOwner("")`, so I probably would rather use explicit `WithOwner`, even with empty owner

pkg/controller/generic/cleanup/cleanup.go
DmitriyMV: Some sort of `combine` should do the trick. I haven't really thought about it.

smira: but this seems tricky right now, as input is singular, but outputs are not singular

pkg/controller/runtime/adapter.go
```diff
@@ -338,11 +338,20 @@ func (adapter *adapter) Teardown(ctx context.Context, resourcePointer resource.P
...
 
-	return adapter.runtime.state.Teardown(ctx, resourcePointer, state.WithTeardownOwner(adapter.name))
+	var opts []state.TeardownOption
+
+	opOpt := controller.ToOptions(opOpts...)
+	if opOpt.Owner != nil {
+		opts = append(opts, state.WithTeardownOwner(*opOpt.Owner))
+	} else {
+		opts = append(opts, state.WithTeardownOwner(adapter.name))
+	}
```
DmitriyMV, line 348: @smira explicitly passing owner now.

pkg/controller/generic/cleanup/cleanup.go
DmitriyMV: Done in `Combine` func.

pkg/controller/generic/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,292 @@
...
+			),
+		},
+	)
+}
+
+func TestNoRemovalUntilNoOutputsMultiple(t *testing.T) {
+	runTest(t, func(ctx context.Context, t *testing.T, st state.State, rt *runtime.Runtime) {
+		ctrl := NewOutputControllerMultiple()
+
+		require.NoError(t, rt.RegisterController(ctrl))
```
DmitriyMV, line 224: @smira multiple outputs watcher

DmitriyMV:
Depends on #249 #250 and https://github.com/cosi-project/specification/pull/8

pkg/controller/generic/cleanup/cleanup.go
```diff
@@ -0,0 +1,431 @@
...
+			}
+
+			var multiErr error
+
+			var inTearDown int
+
+			for iter := safe.IteratorFromList(list); iter.Next(); {
+				out := iter.Value()
+
+				ready, err := r.Teardown(ctx, out.Metadata(), controller.WithOwner(""))
```
smira, line 289: should we skip here resources which have `owner != ""` ? i.e. those resources which should be cleaned up by other controllers?

we should probably still wait on them, but we shouldn't attempt tearing them down, as this will fail?

smira (approved):

PR #250, opened 2023-04-04
feat: add new package testutils

DmitriyMV:
Since `*testing.T` is explicitly not goroutine safe (and there is even `go vet` check), you can't create new goroutines in
test and use `*testing.T`. One approach is to return errors explicitly from tests helpers, but this doesn't scale well. This
commit does another - it provides a safe wrapper around any implementation of `*testing.T` that accepts and returns a context, which will be canceled on any test failure.

As an example - it fixes a small concurrent bug in `setup` function in `transform` package tests.


DmitriyMV:
Depends on #259

pkg/controller/generic/transform/transform_test.go
```diff
@@ -472,13 +470,15 @@ func setup(t *testing.T, f func(ctx context.Context, st state.State, rt *runtime
 
 	wg.Add(1)
 
+	wrappedT, ctx := testutils.WrapT(ctx, t)
+
 	go func() {
 		defer wg.Done()
 
-		assert.NoError(runtime.Run(ctx))
```
smira, line 478: I feel this could be easily rewritten using `errCh`, and we don't need the whole wrappers thing? 

I'm still not sure that things work correctly with `t.Fail()` in another goroutine

smira: `rtestutils` should always be in the "main" goroutine of the test? 

I'm trying to understand the issue better

DmitriyMV: Use case example - you can't use Destroy as as helper for resource destruction from the helper goroutine. If something fails, there is no way to find out. You can go with Teardown - sync - Destroy approach, but this isn't exactly the same flow. You also have to mindful of where you pass `*testing.T`.

To be honest I'm not exactly convinced of this approach either. But the errCh and explicit point of error passing will mean that all our test helpers (functions which accept `T`) are not just goroutine unsafe but should always be called from the testing routine.

The alternative is to rework rtestutils package to return errors, but you'll still need some sort of collector for the callbacks.

PR #252, opened 2023-04-07
chore: pass compression options to `zstd.Encoder`

DmitriyMV:
This commit drastically decreases memory consumption in some cases.

### DIFF
--- a/pkg/state/impl/store/compression/compression_test.go
+++ b/pkg/state/impl/store/compression/compression_test.go
@@ -5,6 +5,9 @@
 package compression_test
 
 import (
+	"math/rand"
+	"runtime"
+	"runtime/debug"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


smira (approved):
should we tie it into the Benchmark? I think I had one

And also might be nice to assert on the compression size.

But overall, great catch! :1st_place_medal:

smira:
I think also part of that (I guess) is that it scales with `GOMAXPROCS`, and it runs on a machine with many cores.

andrewrynhard:
I came here only to say that this is a `fix` ... but I don't expect you to change it here. Just something to keep in mind for other PRs. Chores are usually commits that don't have production code changes and are CI type changes.

andrewrynhard:
> I came here only to say that this is a `fix` ... but I don't expect you to change it here. Just something to keep in mind for other PRs. Chores are usually commits that don't have production code changes and are CI type changes.

Other than that... awesome find!

PR #274, opened 2023-06-04
chore: migrate from sync.Map to xsync.MapOf

DmitriyMV:
See [this](https://puzpuzpuz.dev/so-long-syncmap) and [this](https://github.com/golang/go/issues/47643). Map growing works a bit differently, and it doesn't support nil, but we do not store nils anyway.


frezbo:
Would this eventually be in the go stdlib, the issue doesn't seem to indicate such or I missed it?

DmitriyMV:
> Would this eventually be in the go stdlib, the issue doesn't seem to indicate such or I missed it?

Unlikely at this moment - discussion didn't get much traction in 2021 so author retracted the proposal.

DmitriyMV:
@frezbo created alternative PR in #275. I think, while I still like the idea of replacing sync.Map with something else, I would like to try it first on something not **that** critical.

PR #276, opened 2023-06-04
chore: add type-safe wrapper around sync.Map

DmitriyMV:
Alternative to #274 - this PR adds typesafe wrapper around sync.Map. The main benefit is that we can drop it, once generic version of sync.Map arrives.


smira (approved):
:cool:

DmitriyMV:
Added tests which test type-assertions. Previous version was failing on `nil`s.

PR #298, opened 2023-07-14
chore: bump kres, generate clone method

DmitriyMV:
Add fast-path `CloneVT` type assertion to `protobuf.ResourceSpec`.


pkg/resource/typed/typed_resource_test.go
```diff
@@ -135,3 +146,40 @@ func TestUnmarshalJSON(t *testing.T) {
...
+	)
+
+	benchRes := testing.Benchmark(func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			BenchStore = res.DeepCopy()
+		}
+	})
+
+	// Report ns, allocs and bytes per op
+	t.Logf("ns/op: %d, allocs/op: %d, bytes/op: %d", benchRes.NsPerOp(), benchRes.AllocsPerOp(), benchRes.AllocedBytesPerOp())
```
smira, line 184: should we actually assert on allocations here?

smira (approved):

pkg/resource/typed/typed_resource_test.go
DmitriyMV: Done!

PR #300, opened 2023-07-24
feat: allow TransformController to skip over some inputs

smira:
This allows to map only _some_ Inputs to Outputs.

Also add an automatic lightweight tracker to adapters.

### DIFF
--- a/pkg/controller/generic/transform/controller.go
+++ b/pkg/controller/generic/transform/controller.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/siderolabs/gen/channel"
+	"github.com/siderolabs/gen/optional"
 	"github.com/siderolabs/gen/xerrors"
 	"go.uber.org/zap"
 
@@ -27,7 +28,7 @@ import (
 //     to be fully removed before attempting to delete the output.
 //   - if this controller is configured to set finalizers on its inputs, the finalizer will only be removed when matching output is destroyed.
 type Controller[Input generic.ResourceWithRD, Output generic.ResourceWithRD] struct {
-	mapFunc              func(Input) Output
+	mapFunc              func(Input) optional.Optional[Output]
 	transformFunc        func(context.Context, controller.ReaderWriter, *zap.Logger, Input, Output) error
 	finalizerRemovalFunc func(context.Context, controller.ReaderWriter, *zap.Logger, Input) error
 	generic.NamedController
@@ -42,6 +43,10 @@ type Settings[Input generic.ResourceWithRD, Output generic.ResourceWithRD] struc
 	//
 	// Only Output metadata is important, the spec is ignored.
 	MapMetadataFunc func(Input) Output
+	// MapMetadataOptionalFunc acts like a MapMetadataFunc, but returns optional Output.
+	//
+	// If the Output is not present, the controller will skip the input.
+	MapMetadataOptionalFunc func(Input) optional.Optional[Output]
 	// TransformFunc should modify Output based on Input and any additional resources fetched via Reader.
 	//
 	// If TransformFunc returns error tagged with SkipReconcileTag, the error is ignored and the controller will
@@ -68,6 +73,8 @@ type Settings[Input generic.ResourceWithRD, Output generic.ResourceWithRD] struc
 }
 
 // NewController creates a new TransformController.
+//
+//nolint:gocognit,gocyclo,cyclop
 func NewController[Input generic.ResourceWithRD, Output generic.ResourceWithRD](
 	settings Settings[Input, Output],
 	opts ...ControllerOption,
@@ -79,8 +86,10 @@ func NewController[Input generic.ResourceWithRD, Output generic.ResourceWithRD](
 	}
 
 	switch {
-	case settings.MapMetadataFunc == nil:
-		panic("MapFunc is required")
+	case settings.MapMetadataFunc == nil && settings.MapMetadataOptionalFunc == nil:
+		panic("MapMetadataFunc is required")
+	case settings.MapMetadataFunc != nil && settings.MapMetadataOptionalFunc != nil:
+		panic("MapMetadataFunc and MapMetadataOptionalFunc are mutually exclusive")
 	case settings.TransformFunc == nil && len(options.extraOutputs) == 0:
 		panic("TransformFunc is required")
 	case settings.TransformExtraOutputFunc == nil && len(options.extraOutputs) > 0:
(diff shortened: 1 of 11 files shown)


pkg/controller/generic/transform/controller.go
```diff
@@ -241,6 +245,14 @@ func (ctrl *Controller[Input, Output]) processInputs(
 
 		mappedOut := ctrl.mapFunc(in)
 
+		if mappedOut == zeroOutput {
```
DmitriyMV, line 248: I think we should go with something like `func isZero[T any](val T) bool { reflect.ValueOf(v).IsZero() }`. I know that using reflect here is not ideal, but it will bring almost no performance penalty for small objects (smaller than 1000 bytes) and we don't need to add new `ComparableResourceWithRD` constraint. On my test IsZero for small objects is around 4-8ns.

DmitriyMV: Actually, thinking this further - maybe we should just use `MapMetadataFunc func(Input) (Output, bool)` or `func(Input) Optional[Output]`?

smira: zero here is just `nil`, I don't see why?

DmitriyMV: Yeah, but technically `Controller` make no assumptions about that, that why you can't use `nil` directly. At the same time using `comparable` here looks like a bit of hack, since you are comparing stuff with zero, and not with each other.

smira: not sure I get it. all resources from practical point of view are pointers to some type, so zero value is nil, but as I can't use `nil` directly, I use zero value which we use almost everywhere else. so it compiles to `if a == nil`, and that is what we want, rather than going via reflection for real

smira: we could do `Optional` as well, but that would break the API

smira: reworked to be optional

Unix4ever (approved):

smira:
Example of boilerplate removal (in Talos):

```diff
--- a/internal/app/machined/pkg/controllers/cri/seccomp_profile.go
+++ b/internal/app/machined/pkg/controllers/cri/seccomp_profile.go
@@ -68,7 +68,7 @@ func (ctrl *SeccompProfileController) Run(ctx context.Context, r controller.Runt
                        return fmt.Errorf("error getting config: %w", err)
                }
 
-               touchedIDs := map[string]struct{}{}
+               r.StartTrackingOutputs()
 
                if cfg.Config().Machine() != nil {
                        for _, profile := range cfg.Config().Machine().SeccompProfiles() {
@@ -80,27 +80,9 @@ func (ctrl *SeccompProfileController) Run(ctx context.Context, r controller.Runt
                                }); err != nil {
                                        return err
                                }
-
-                               touchedIDs[profile.Name()] = struct{}{}
-                       }
-               }
-
-               // list keys for cleanup
-               list, err := safe.ReaderListAll[*cri.SeccompProfile](ctx, r)
-               if err != nil {
-                       return fmt.Errorf("error listing seccomp profiles: %w", err)
-               }
-
-               for iter := safe.IteratorFromList(list); iter.Next(); {
-                       profile := iter.Value()
-
-                       if _, ok := touchedIDs[profile.Metadata().ID()]; !ok {
-                               if err := r.Destroy(ctx, profile.Metadata()); err != nil {
-                                       return fmt.Errorf("error deleting seccomp profile: %w", err)
-                               }
                        }
                }
 
-               r.ResetRestartBackoff()
+               safe.CleanupOutputs[*cri.SeccompProfile](ctx, r)
        }
 }
```

pkg/controller/runtime/adapter.go
```diff
@@ -479,3 +499,52 @@ func (adapter *adapter) runOnce(ctx context.Context, logger *zap.Logger) (err er
 
 	return adapter.ctrl.Run(ctx, adapter, logger)
 }
+
+func outputTrackingID(md *resource.Metadata) string {
+	return fmt.Sprintf("%s/%s/%s", md.Namespace(), md.Type(), md.ID())
+}
```
DmitriyMV, line 505: Why string tho? Can't we return custom struct or use `ordered.Triple` from `gen`? This will remove unnecessary allocations

pkg/controller/runtime/adapter.go
```diff
@@ -479,3 +499,52 @@ func (adapter *adapter) runOnce(ctx context.Context, logger *zap.Logger) (err er
...
+				continue
+			}
+
+			if err = adapter.Destroy(ctx, resource.Metadata()); err != nil {
+				return fmt.Errorf("error destroying resource %s: %w", resource.Metadata(), err)
+			}
+		}
+	}
+
+	adapter.outputTracker = nil
```
DmitriyMV, line 545: We can preserve map capacity here:
```suggestion
	for k := range adapter.outputTracker {
		delete(adapter.outputTracker, k)
	}
```
This will be compiled `map_clear` I think.

DmitriyMV: Or! Add global pool and use it. That way we can preserve check in `StartTrackingOutputs`

pkg/controller/runtime/adapter.go
smira: good idea, I will rework

pkg/controller/runtime/adapter.go
smira: no, I need explicit `nil` to make sure the calls are always properly paired: `StartTrackingOutputs()` & `CleanupOutputs()`, `nil` is an explicit signal here.

pkg/controller/runtime/adapter.go
smira: updated

pkg/controller/runtime/adapter.go
smira: added a sync.Pool, I somehow didn't see your sync.Pool comment first :-D

pkg/controller/runtime/runtime.go
```diff
@@ -36,6 +36,8 @@ type Runtime struct { //nolint:govet
 	watchedMu   sync.Mutex
 	watched     map[watchKey]struct{}
 
+	trackingOutputPool trackingOutputPool
```
DmitriyMV, line 39: I would actually make it a global variable, since `sync.Pool` is thread safe. The reason for making it a global variable - it will amortize allocations across all controllers as opposed to being "per controller allocator".

pkg/controller/generic/transform/controller.go
```diff
@@ -239,7 +255,12 @@ func (ctrl *Controller[Input, Output]) processInputs(
 	for iter := safe.IteratorFromList(inputItems); iter.Next(); {
 		in := iter.Value()
 
-		mappedOut := ctrl.mapFunc(in)
+		mappedOut, present := ctrl.mapFunc(in).Get()
+
```
DmitriyMV, line 259: Empty line?

pkg/controller/protobuf/client/client.go
```diff
@@ -721,6 +721,14 @@ func (ctrlAdapter *controllerAdapter) RemoveFinalizer(ctx context.Context, resou
...
 }
 
+func (ctrlAdapter *controllerAdapter) StartTrackingOutputs() {
+	panic("not implemented")
+}
+
+func (ctrlAdapter *controllerAdapter) CleanupOutputs(context.Context, ...resource.Kind) error {
+	panic("not implemented")
+}
+
```
DmitriyMV, line 731: I think it will be better if we write here something like `panic("remote controller output tracking is not supported")` instead of generic "not implemented".

DmitriyMV (approved):
Minor suggestions, but other than that 🆒

pkg/controller/runtime/runtime.go
smira: this is not per-controller! it's per-controller runtime

smira: I would prefer to keep runtimes isolated, and no global vars. Usual case is to have a single controller-runtime per process, so it's effectively shared pool.

PR #304, opened 2023-08-02
chore: add `ReconcileEventCh` support type

DmitriyMV:
By defining custom type on top of `<-chan ReconcileEvent` and defining helper method, we can write code like this:

```go
if !r.EventCh().Recv(ctx) {
	return nil
}
```

Which is more straightforward than

```go
select {
case <-ctx.Done():
	return false
case <-r.EventCh():
	// our logic here
}
```

And definitely easier than:

```go
if _, ok := channel.RecvWithContext(ctx, r.EventCh()); !ok && ctx.Err() != nil {
	return nil //nolint:nilerr
}
```

This is a breaking change, but in reality it should not require any updates to Talos code.

### DIFF
--- a/pkg/controller/conformance/controllers.go
+++ b/pkg/controller/conformance/controllers.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"strconv"
 
-	"github.com/siderolabs/gen/channel"
 	"go.uber.org/zap"
 
 	"github.com/cosi-project/runtime/pkg/controller"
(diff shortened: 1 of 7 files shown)


DmitriyMV:
Related https://github.com/siderolabs/gen/pull/13

smira:
The original idea I had was to have a channel, as it can be combined in a single `select` various ways, the easiest is the one you have in the example (just waiting for EventCh), but there are many cases when we combine it with a `time.NewTimer`, or some external event source (watching Linux kernel events), internal controller source (over some other channel). 

I would rather think about refactoring the `Run` method into an eaiser form of for those controllers which don't care about other event sources, and just peek on the `EventCh` in an endless loop.

Something like you can have `Run(...)` which is full implementation, and you have `EasyRun(...)` which gets called once per `EventCh` receive if that makes sense.

DmitriyMV:
@smira we can go two ways:

Create a wrapper for controller, which is essentially what `transform.Controller` is under the hood. Or we can create separate `RegisterEasyController` function which will do pretty much the same.

smira:
I feel like RegisterSimpleController might be good with automatic wrapping, but also I think we need to see how much it would actually help, before we jump into implementation.

PR #309, opened 2023-08-17
feat: introduce new query operators for label selectors

Unix4ever:
New operators:

- `IN` - looks up the value in the set of values.
- `LT` - less than (string comparison).
- `LTE` - less than equal (string comparison).
- `LT_NUMERIC` - less than (numeric comparison, tries to parse humanized values).
- `LTE_NUMERIC` - less than equal (numeric comparison, tries to parse humanized values).

All terms can now be inverted by the new field in the spec: `Invert`. For options it will be `resource.LabelIn("a", []string{"b", "c"}, resource.NotMatches)`.

Dropped `NOT_EXISTS` in favor of new `Invert` flag.


Unix4ever:
Depends on https://github.com/cosi-project/specification/pull/9

.golangci.yml
```diff
@@ -74,6 +74,8 @@ linters-settings:
   govet:
     check-shadowing: true
     enable-all: true
+    disable:
```
smira, line 77: should bump `go.mod` to `go 1.21`

pkg/controller/protobuf/client/client.go
```diff
@@ -392,19 +392,51 @@ func (ctrlAdapter *controllerAdapter) List(ctx context.Context, resourceKind res
 			switch term.Op {
 			case resource.LabelOpEqual:
 				labelQuery.Terms = append(labelQuery.Terms, &v1alpha1.LabelTerm{
-					Key:   term.Key,
```
smira, line 395: I think there's a conversion in opposite direction in the `protobuf/`, also you should use new expression in conformance tests for COSI state

pkg/resource/labels.go
```diff
@@ -27,28 +32,130 @@ func (labels Labels) Equal(other Labels) bool {
...
 
-		return value == term.Value
+		return left <= right
 	default:
 		panic(fmt.Sprintf("unsupported label term operator: %v", term.Op))
 	}
 }
+
+var (
+	re    = regexp.MustCompile(`^(\d+)(.*)$`)
```
smira, line 86: let's not use global variables as `regexp`, this causes allocations on import even if you never use it. and I think this can be easily parsed with a straightfowards Go function which will runs faster

pkg/resource/labels.go
```diff
@@ -27,28 +32,130 @@ func (labels Labels) Equal(other Labels) bool {
...
+
+	numRight, ok := parseValue(right)
+	if !ok {
+		return 0, 0, false
+	}
+
+	return numLeft, numRight, true
+}
+
+func parseValue(value string) (int64, bool) {
```
smira, line 116: it will be cool to extract this function and related to a small `internal/` sub-package, and have it covered with proper tests

pkg/resource/internal/compare/compare.go
```diff
@@ -0,0 +1,108 @@
...
+	case "k":
+		val = 1e3
+	default:
+		return 0, false
+	}
+
+	return val, true
+}
+
+func getMultiplier(value string) int64 {
```
smira, line 90: I think this can be simplified a bit by simply doing a match on `value[:2]` if the len > 2, and then for `value[:1]` if the previous one never matched? 

as the function `getMultiplierForUnits` is already explicitly matching for either one-character or two-character prefix?

pkg/resource/internal/compare/compare.go
```diff
@@ -0,0 +1,108 @@
...
+
+func parseValue(value string) (int64, bool) {
+	value = strings.TrimSpace(value)
+
+	var (
+		digits string
+		rest   string
+	)
+
+	for _, letter := range value {
```
smira, line 37: we could do better job here by finding the first `i` in the string which is non-digit, and splitting the string on that index, we don't need to build it chracter-by-character

pkg/resource/internal/compare/compare.go
```diff
@@ -0,0 +1,108 @@
...
+		if rest == "" && unicode.IsDigit(letter) || string(letter) == "." {
```
smira, line 38: why `.` if we parse as `int64`?

if we parse as `int64`, why don't we allow `-33` ?

pkg/resource/internal/compare/compare.go
```diff
@@ -0,0 +1,108 @@
...
+	if len(digits) == 0 {
+		return 0, false
+	}
+
+	res, err := strconv.ParseInt(digits, 10, 64)
+	if err != nil {
+		return 0, false
+	}
+
+	return res * getMultiplier(rest), true
```
smira, line 56: not clear if we allow invalid `multiplier` here? should it be documented? 

e.g. `1x` parses as `1` as it seems

Unix4ever: I guess we can just disable invalid `multipliers` here and it will also solve this `.` issue as well.

pkg/resource/internal/compare/compare.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return numLeft, numRight, true
+}
+
+func parseValue(value string) (int64, bool) {
+	value = strings.TrimSpace(value)
+
+	sign := int64(1)
+
+	if string(value[0]) == "-" {
```
smira, line 34: `strconv.ParseInt` handles negative numbers perfectly fine, I don't think we need to handle it this way

pkg/resource/internal/compare/compare.go
```diff
@@ -0,0 +1,117 @@
...
+		value = value[1:]
+		sign = -1
+	}
+
+	splitPoint := len(value)
+
+	for i, letter := range value {
+		if !unicode.IsDigit(letter) {
```
smira, line 42: this will handle many more characters as 'digits', including those `ParseInt` can't handle

a proper check might be `c >= '0' && c <= '9'`

pkg/resource/internal/compare/compare.go
```diff
@@ -0,0 +1,117 @@
...
+		val = 1e9
+	case "m":
+		val = 1e6
+	case "k":
+		val = 1e3
+	default:
+		return 0, false
+	}
+
+	return val, true
```
smira, line 96: why? if we have `default:` ?

Unix4ever: the code above doesn't do return, so this is necessary.
But I'll update the code as you suggested so it'll be irrelevant anyways

smira (approved):

PR #312, opened 2023-08-24
feat: add support for passing multiple label queries in watch and list

Unix4ever:
Terms inside query are still handled with `AND`, while separate label queries are handled with `OR`.


smira (approved):

pkg/state/conformance/state.go
```diff
@@ -453,6 +454,18 @@ func (suite *StateSuite) testWatchKindWithLabels(useAggregated bool) {
 		state.WatchWithLabelQuery(resource.LabelExists(labelCommon)),
 	))
 
+	// watch with label == label2 || label == label3
+	suite.Require().NoError(watchAggregateAdapter(
```
frezbo, line 458: so watch returns if only `label == label3` is only matched?

frezbo (approved):

pkg/state/conformance/state.go
Unix4ever: hmm, not sure I've got the question :)
But the idea is that if you provide two label query options and either of them matches we add the item to the watch.

So the list will contain both res with `label = label2` and `label = label3`

PR #315, opened 2023-08-28
chore: update stuff

DmitriyMV:
- Bump ci script.
- Bump github.com/ProtonMail/gopenpgp/v2 to v2.7.3
- Bump github.com/grpc-ecosystem/grpc-gateway/v2 to v2.17.1
- Add `List.Iterator` method and deprecate `IteratorFromList`.
- `safe.State*` functions should accept `state.CoreState` where possible.


pkg/safe/state.go
```diff
@@ -299,13 +299,20 @@ func (l *List[T]) Find(fn func(T) bool) (T, bool) {
 	return zero, false
 }
 
+// Iterator returns a new iterator over the list.
+func (l *List[T]) Iterator() ListIterator[T] {
```
frezbo, line 303: any particular reason to deprecate, or just for easier use?

frezbo (approved):

pkg/safe/state.go
DmitriyMV: Just easier to use. I don't remember why I didn't do this in the first place. It looks better too!

frezbo: yes, feels better than having to call safe.Iterator

PR #318, opened 2023-08-31
fix: properly handle invert for comparison, when key not found

Unix4ever:
Should return `nil` if the key doesn't exist in the labels.

### DIFF
--- a/pkg/resource/labels.go
+++ b/pkg/resource/labels.go
@@ -54,6 +54,10 @@ func (labels Labels) matches(term LabelTerm) *bool {
 	value, ok := labels.Get(term.Key)
 
 	if !ok {
+		if term.Op.isComparison() {
+			return nil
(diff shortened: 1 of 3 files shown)


pkg/resource/labels.go
```diff
@@ -54,6 +54,10 @@ func (labels Labels) matches(term LabelTerm) *bool {
 	value, ok := labels.Get(term.Key)
 
 	if !ok {
+		if term.Op > LabelOpIn {
```
smira, line 57: isn't this too fragile?

smira (approved):

frezbo (approved):

pkg/resource/labels.go
Unix4ever: it is, can make it more strick

PR #322, opened 2023-09-04
feat: add extra channel support to transform controller

utkuozdemir:
Add an option to specify an extra channel to the transform controller. Controller will wake up upon receiving data from the channel and run transform.

### DIFF
--- a/pkg/controller/generic/transform/controller.go
+++ b/pkg/controller/generic/transform/controller.go
@@ -198,6 +198,7 @@ func (ctrl *Controller[Input, Output]) Run(ctx context.Context, r controller.Run
 		case <-ctx.Done():
 			return nil
 		case <-r.EventCh():
+		case <-ctrl.options.extraEventCh:
 		}
 
 		// controller runs in two phases:
--- a/pkg/controller/generic/transform/options.go
+++ b/pkg/controller/generic/transform/options.go
@@ -11,6 +11,7 @@ import (
 
 // ControllerOptions configures TransformController.
 type ControllerOptions struct {
+	extraEventCh            <-chan struct{}
 	inputListOptions        []state.ListOption
 	extraInputs             []controller.Input
 	extraOutputs            []controller.Output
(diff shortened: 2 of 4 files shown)


pkg/controller/generic/transform/controller.go
```diff
@@ -198,6 +198,7 @@ func (ctrl *Controller[Input, Output]) Run(ctx context.Context, r controller.Run
 		case <-ctx.Done():
 			return nil
 		case <-r.EventCh():
+		case <-ctrl.options.extraCh:
```
frezbo, line 201: maybe call extraChInput?

pkg/controller/generic/transform/resource_test.go
```diff
@@ -79,7 +79,8 @@ func (BE) ResourceDefinition() meta.ResourceDefinitionSpec {
 
 // BSpec provides B definition.
 type BSpec struct {
-	Out string
+	Out            string
+	TransformCount int
```
frezbo, line 83: don;t see code incrementing it?

pkg/controller/generic/transform/options.go
```diff
@@ -70,3 +71,13 @@ func WithIgnoreTearingDownInputs() ControllerOption {
 		o.ignoreTearingDownInputs = true
 	}
 }
+
+// WithExtraChannel adds an extra channel to the controller.
+//
+// When this channel receives data, the controller will run the transform function.
+// This is useful to wake up the controller from a goroutine.
+func WithExtraChannel(extraCh <-chan struct{}) ControllerOption {
```
smira, line 79:
```suggestion
func WithExtraEventChannel(extraEventCh <-chan struct{}) ControllerOption {
```

smira (approved):

pkg/controller/generic/transform/controller.go
frezbo: just the variable name :sweat_smile:

frezbo (approved):

PR #323, opened 2023-09-04
feat: add on shutdown callback to the transform controller

utkuozdemir:
Add a callback option to the transform controller which is run just before the controller is being shut down gracefully. This allows a transform controller to do cleanups, e.g. waiting for goroutines it spawned to complete.


Unix4ever (approved):

pkg/controller/generic/transform/controller.go
```diff
@@ -196,6 +196,10 @@ func (ctrl *Controller[Input, Output]) Run(ctx context.Context, r controller.Run
 	for {
 		select {
 		case <-ctx.Done():
+			if ctrl.options.onShutdownCallback != nil {
```
smira, line 199: this should better be a `defer`

pkg/controller/generic/transform/transform_test.go
```diff
@@ -537,6 +562,10 @@ func setup(t *testing.T, f func(ctx context.Context, st state.State, rt *runtime
 		if err != nil && !errors.Is(err, context.Canceled) {
 			t.Fatal(err)
 		}
+
+		if onShutdown != nil {
```
smira, line 566: you could attach this via `t.Cleanup()` probably in the test itself, no need to pass it through here

frezbo (approved):

pkg/controller/generic/transform/controller.go
smira: I don't think that discussion has anything about _successful_ shutdown. On any exit you want to make sure that all goroutines/workers terminated.

Not only that, you have to use a context which is bound to the controller execution, so if the controller aborts on any reason, context is canceled.

DmitriyMV: Maybe we can do something like this:
```go
ctx, cancel := context.WithCancel(ctx)
defer cancel()

stop := context.AfterFunc(ctx, func() {
	if ctrl.options.onShutdownCallback != nil {
		ctrl.options.onShutdownCallback()
	}
})
defer stop()
```

?

smira: hmm... can't we just `defer ...onShutdownCallback()` before `defer cancel()` to make sure it gets called after the context is canceled?

utkuozdemir: I think it would work fine ⬆️ 

Wdyt about allowing to return error from this callback overall? Should I ditch it or keep (despite being in defer)?

smira: I feel such callbacks might get a logger as an argument, but reporting an error feels like too late (as the controller is already terminating either as a result of error or as the runtime is terminating).

also I wonder if we could use `goleak` in the unit-tests as a defence in depth: make sure controller terminates and cleans up all goroutines

smira: (I mean `goleak` in Omni, runtime tests here should be already using it)

utkuozdemir: @smira @DmitriyMV I (think I) did the changes. Could you guys have another look to it?

pkg/controller/generic/transform/controller.go
```diff
@@ -187,12 +187,22 @@ type runState struct {
 }
 
 // Run implements controller.Controller interface.
-func (ctrl *Controller[Input, Output]) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
```
smira, line 190: hopefully named returned arg can be removed now :see_no_evil:

smira (approved):

pkg/controller/generic/transform/options.go
```diff
@@ -5,13 +5,16 @@
...
+	"go.uber.org/zap"
+
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/state"
 )
 
 // ControllerOptions configures TransformController.
 type ControllerOptions struct {
 	extraEventCh            <-chan struct{}
+	onShutdownCallback      func(logger *zap.Logger)
```
DmitriyMV, line 17: Maybe it makes sense to pass `controller.ReaderWriter` here?

pkg/controller/generic/transform/options.go
```diff
@@ -81,3 +84,10 @@ func WithExtraEventChannel(extraEventCh <-chan struct{}) ControllerOption {
 		o.extraEventCh = extraEventCh
 	}
 }
+
+// WithOnShutdownCallback adds a callback to be called when the controller is shutting down, either gracefully or due to an error.
+func WithOnShutdownCallback(onShutdownCallback func(logger *zap.Logger)) ControllerOption {
```
DmitriyMV, line 89: Explicit signature for `func(logger *zap.Logger)` would be nice.

pkg/controller/generic/transform/options.go
smira: I don't care, whatever works better

PR #325, opened 2023-09-07
feat: support `OutputShared` for the main res in Transform controller

Unix4ever:
This will provide more flexibility in the transform controller usage.

### DIFF
--- a/pkg/controller/generic/transform/options.go
+++ b/pkg/controller/generic/transform/options.go
@@ -20,6 +20,7 @@ type ControllerOptions struct {
 	inputListOptions        []state.ListOption
 	extraInputs             []controller.Input
 	extraOutputs            []controller.Output
+	primaryOutputKind       controller.OutputKind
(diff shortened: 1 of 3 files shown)


pkg/controller/generic/transform/options.go
```diff
@@ -96,3 +97,10 @@ func WithOnShutdownCallback(onShutdownCallback OnShutdownCallback) ControllerOpt
 		o.onShutdownCallback = onShutdownCallback
 	}
 }
+
+// WithOutputKind sets main output resource kind.
+func WithOutputKind(kind controller.OutputKind) ControllerOption {
```
frezbo, line 102: is the default still `OutputExclusive`?

Unix4ever: `OutputExclusive` is 0, so when unset it's going to default to it.
But I can set it explicitly, let me change that.

smira (approved):

PR #333, opened 2023-09-18
chore: bump deps

DmitriyMV:
- run rekres
- use `slices` and `xslices` module where possible
- bump actions/checkout to v4.0.0
- bump docker/setup-qemu-action to v3.0.0
- bump docker/setup-buildx-action to v3.0.0
- bump docker/login-action to v3.0.0
- bump github.com/grpc-ecosystem/grpc-gateway/v2 to v2.18.0
- bump github.com/siderolabs/gen to v0.4.7
- bump go.uber.org/zap to v1.26.0
- bump google.golang.org/grpc to v1.58.1
- Use `clear` builtin instead of manual deletion from map using range.


smira:
we don't need to use new GHA generator here probably, disable it and keep using "Drone" for now

frezbo:
> we don't need to use new GHA generator here probably, disable it and keep using "Drone" for now

this already use gh actions I believe? :thinking:

smira:
> > we don't need to use new GHA generator here probably, disable it and keep using "Drone" for now
> 
> this already use gh actions I believe? 🤔

yes, this one lives under a different org, so no need to mix & match everything, we can keep using existing working flow for now

DmitriyMV:
@smira done!

frezbo (approved):

PR #335, opened 2023-09-22
refactor: optimize the metadata.Equal for small number of finalizers

### LINKED ISSUE #306: optimize metadata.Equal
the `sort.Strings()` and a copy of finalizers - can we do a better version? should we optimize for a special case of 1,2 items?

smira:
It might be pretty common to have 1-2 finalizers, so we can do a special case for it.

```
name                old time/op    new time/op    delta
MetadataEqual/0-32    12.1ns ± 2%    11.8ns ± 2%      ~     (p=0.057 n=4+4)
MetadataEqual/1-32     206ns ± 8%      14ns ± 0%   -93.11%  (p=0.029 n=4+4)
MetadataEqual/2-32     268ns ± 7%      21ns ± 2%   -92.13%  (p=0.029 n=4+4)
MetadataEqual/3-32     328ns ± 4%     227ns ± 4%   -30.86%  (p=0.029 n=4+4)

name                old alloc/op   new alloc/op   delta
MetadataEqual/0-32     0.00B          0.00B           ~     (all equal)
MetadataEqual/1-32     80.0B ± 0%      0.0B       -100.00%  (p=0.029 n=4+4)
MetadataEqual/2-32      112B ± 0%        0B       -100.00%  (p=0.029 n=4+4)
MetadataEqual/3-32      144B ± 0%       96B ± 0%   -33.33%  (p=0.029 n=4+4)

name                old allocs/op  new allocs/op  delta
MetadataEqual/0-32      0.00           0.00           ~     (all equal)
MetadataEqual/1-32      4.00 ± 0%      0.00       -100.00%  (p=0.029 n=4+4)
MetadataEqual/2-32      4.00 ± 0%      0.00       -100.00%  (p=0.029 n=4+4)
MetadataEqual/3-32      4.00 ± 0%      2.00 ± 0%   -50.00%  (p=0.029 n=4+4)
```

Fixes #306

### DIFF
--- a/pkg/resource/metadata.go
+++ b/pkg/resource/metadata.go
@@ -6,7 +6,7 @@ package resource
 
 import (
 	"fmt"
-	"sort"
+	"slices"
 	"time"
 
 	"github.com/siderolabs/gen/xslices"
@@ -172,19 +172,23 @@ func (md Metadata) Equal(other Metadata) bool {
 		return true
 	}
 
-	fins := append(Finalizers(nil), md.fins...)
-	otherFins := append(Finalizers(nil), other.fins...)
+	switch len(md.fins) {
+	case 0:
+		return true
+	case 1:
+		return md.fins[0] == other.fins[0]
+	case 2:
+		return md.fins[0] == other.fins[0] && md.fins[1] == other.fins[1] ||
+			md.fins[0] == other.fins[1] && md.fins[1] == other.fins[0]
+	default:
+		fins := slices.Clone(md.fins)
+		otherFins := slices.Clone(other.fins)
 
-	sort.Strings(fins)
-	sort.Strings(otherFins)
+		slices.Sort(fins)
+		slices.Sort(otherFins)
 
-	for i := range fins {
-		if fins[i] != otherFins[i] {
-			return false
-		}
+		return slices.Equal(fins, otherFins)
 	}
-
-	return true
 }
 
 // MarshalYAML implements yaml.Marshaller interface.
--- a/pkg/resource/metadata_test.go
+++ b/pkg/resource/metadata_test.go
@@ -6,6 +6,7 @@ package resource_test
 
 import (
 	"fmt"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/resource/metadata.go
```diff
@@ -172,19 +175,29 @@ func (md Metadata) Equal(other Metadata) bool {
...
+		return md.fins[0] == other.fins[0]
+	case 2:
+		return md.fins[0] == other.fins[0] && md.fins[1] == other.fins[1] ||
+			md.fins[0] == other.fins[1] && md.fins[1] == other.fins[0]
+	default:
+		fins := slices.Clone(md.fins)
+		otherFins := slices.Clone(other.fins)
+
+		sort.Strings(fins)
+		sort.Strings(otherFins)
```
DmitriyMV, line 191:
```suggestion
		slices.Sort(fins)
		slices.Sort(otherFins)
```

Generic version of sort should be faster.

pkg/resource/metadata.go
```diff
@@ -172,19 +175,29 @@ func (md Metadata) Equal(other Metadata) bool {
...
+
+		for i := range fins {
+			if fins[i] != otherFins[i] {
+				return false
+			}
```
DmitriyMV, line 196: You can use `slices.Equal` here.

DmitriyMV (approved):

DmitriyMV:
Small idea - we can actually speed up cases up to 16 (after 16 elements ns/op decrease drops below 50%).

```go
func equalUpTo16(a, b []Finalizer) bool {
	if len(a) != len(b) {
		return false
	}

	if len(a) > 16 {
		panic("len(a) is greater than 16")
	}

	var elementCompared uint16

	for i := range a {
		idx := -1

		for j := range b {
			if elementCompared& (1 << j) == 0 && a[i] == b[j] {
				idx = j

				break
			}
		}

		if idx == -1 {
			return false
		}

		elementCompared |= 1 << idx
	}

	return true
}
```

And use it like 
```go
if len(fins) <= 16 {
	equalUpTo16(fins, otherFins)
}
```

pkg/resource/metadata_test.go
```diff
@@ -254,3 +255,25 @@ func TestNewMedataFromProto(t *testing.T) {
 
 	assert.True(t, md.Equal(other))
 }
+
+func BenchmarkMetadataEqual(b *testing.B) {
```
frezbo, line 259: do we have a before and after?

smira: please see the commit message - there's comparison there.

frezbo: sorry, missed that

frezbo (approved):

smira:
> Small idea - we can actually speed up cases up to 16 (after 16 elements ns/op decrease drops below 50%).

I feel this is a bit of overkill? I mean readability vs. optimization :)

DmitriyMV:
> I feel this is a bit of overkill? I mean readability vs. optimization :)

I mean, probably? Depends on how often we see resources with 3-16 finalizers. Interestingly enough `equalUpTo16` is almost as fast for cases of 2 and 3 elements as current direct comparison.

What is more interesting `equalUpTo64` is 20% slower than current implementation with Clone,Sort,Equal.

pkg/resource/metadata.go
smira: updated

pkg/resource/metadata.go
smira: updated

smira:
Updated with @DmitriyMV suggestions, I feel 1-2 finalizers version is good enough, and also updated >= 3 finalizers version is better than before.

PR #336, opened 2023-09-22
refactor: rewrite dependency database

### LINKED ISSUE #305: rework the controller dependency database
The original implementation is using `go-memdb`, which is an overkill given that we know what the structure is, and what are the requests against it.

We should look into optimizing `GetDependentControllers`, as this gets called in a tight loop.

smira:
Drop `go-memdb` and use a hand-written optimized version for our particular usecase.

Breaking change: replace `*string` as controller.Input.ID with `optional.Optional` for readability, comparison and ease of use.

Fixes #305


smira:
Rough comparison:

```
name                        old time/op    new time/op    delta
GetDependentControllers-32    1.11µs ± 2%    0.09µs ±10%  -91.57%  (p=0.029 n=4+4)
BuildDatabase-32              73.0µs ± 5%    90.9µs ± 4%  +24.52%  (p=0.029 n=4+4)

name                        old alloc/op   new alloc/op   delta
GetDependentControllers-32      464B ± 0%       32B ± 0%  -93.10%  (p=0.029 n=4+4)
BuildDatabase-32              48.3kB ± 1%     1.2kB ± 3%  -97.43%  (p=0.029 n=4+4)

name                        old allocs/op  new allocs/op  delta
GetDependentControllers-32      20.0 ± 0%       1.0 ± 0%  -95.00%  (p=0.029 n=4+4)
BuildDatabase-32                 850 ± 1%         7 ± 0%  -99.18%  (p=0.029 n=4+4)
```

frezbo (approved):

pkg/controller/runtime/dependency/dependency.go
```diff
@@ -6,45 +6,48 @@
...
 )
 
 // Less compares two controller.Dependency objects.
 //
 // This sort order is compatible with the way memdb handles ordering.
-func Less(a, b *controller.Input) bool {
-	aID := StarID
-	if a.ID != nil {
-		aID = *a.ID
+func Less(a, b controller.Input) bool {
```
smira, line 15: need to look into this and the way we return the inputs

smira:
Updated benchmark (benchmark itself slightly changed as well):

```
name                        old time/op    new time/op    delta
GetDependentControllers-32    1.08µs ± 4%    0.11µs ± 5%  -90.08%  (p=0.029 n=4+4)
BuildDatabase-32              71.8µs ± 6%     6.1µs ± 2%  -91.45%  (p=0.029 n=4+4)

name                        old alloc/op   new alloc/op   delta
GetDependentControllers-32      464B ± 0%       32B ± 0%  -93.10%  (p=0.029 n=4+4)
BuildDatabase-32              47.9kB ± 1%     1.3kB ± 8%  -97.30%  (p=0.029 n=4+4)

name                        old allocs/op  new allocs/op  delta
GetDependentControllers-32      20.0 ± 0%       1.0 ± 0%  -95.00%  (p=0.029 n=4+4)
BuildDatabase-32                 843 ± 1%         8 ± 0%  -99.05%  (p=0.029 n=4+4)
```

PR #340, opened 2023-09-28
test: throttle down rtestutils reports

smira:
Instead of reporting on every watch event, report every 30 seconds.


frezbo (approved):

smira:
Plus: don't show duplicate reports.

PR #353, opened 2023-10-20
feat: support requeue in the transform controller

Unix4ever:
This should allow graceful restarts of the controller. For the cases when there is no way to trigger the controller on resource changes.

### DIFF
--- a/pkg/controller/generic/transform/controller.go
+++ b/pkg/controller/generic/transform/controller.go
@@ -7,6 +7,7 @@ package transform
 import (
 	"context"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/generic/transform/controller.go
```diff
@@ -235,6 +256,21 @@ func (ctrl *Controller[Input, Output]) Run(ctx context.Context, r controller.Run
 			return err
 		}
 
+		if state.requeue && ctrl.options.requeueInterval != 0 {
```
smira, line 259: we could in theory build a proper requeue channel with all bells and whistles, for a specific ID if you'd like to?

PR #357, opened 2023-10-25
feat: rework controller read/write metrics

utkuozdemir:
Instead of incrementing the metrics on each call to read or write, wrap the underlying state and increment the metric counters only when a `CoreState` operation succeeds.

Make it possible to disable metrics in runtime.

### DIFF
--- a/pkg/controller/runtime/metrics/state.go
+++ b/pkg/controller/runtime/metrics/state.go
@@ -1,0 +1,73 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package metrics
+
+import (
+	"context"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+)
+
+type metricsWrapper struct {
+	innerState     state.CoreState
+	controllerName string
+}
+
+func (m *metricsWrapper) Get(ctx context.Context, pointer resource.Pointer, option ...state.GetOption) (resource.Resource, error) {
+	ControllerReads.Add(m.controllerName, 1)
+
+	return m.innerState.Get(ctx, pointer, option...)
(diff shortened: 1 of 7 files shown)


utkuozdemir:
@smira I'll add some tests if you approve this way of solving it.

pkg/controller/runtime/metrics/state.go
```diff
@@ -0,0 +1,107 @@
...
+	innerState     state.CoreState
+}
+
+func (m *metricsWrapper) Get(ctx context.Context, pointer resource.Pointer, option ...state.GetOption) (resource.Resource, error) {
+	res, err := m.innerState.Get(ctx, pointer, option...)
+	if err != nil {
+		return nil, err
+	}
+
+	ControllerReads.Add(m.controllerName, 1)
```
smira, line 25: should we move this and other to before the call? if the call fails, the read still happened? 

e.g. if we receive not found, it doesn't mean we haven't read?

smira (approved):

pkg/controller/runtime/metrics/state.go
utkuozdemir: that's tricky - we could do all kinds of checks and error might be from a read/write attempt, or before.

But if the error is `nil`, we know for sure that the operation has actually happened. So I put them to be after the operation, to get no false positives.

If you prefer it to be the other way, I can change it.

smira: I don't think these underlying calls do that many checks... of course it depends, but ignoring e.g. NotFound might lead to some fun stuff

in your previous implementation, you were mis-calculating a lot with `Modify`, this should be more accurate as it will reach `Update` only when there's an update

utkuozdemir: > in your previous implementation, you were mis-calculating a lot with `Modify`, this should be more accurate as it will reach `Update` only when there's an update

Yes, that's why I wrap the CoreState now - much more accurate. It'll also record the Gets happening during Modify etc.

I now also changed it to record metrics regardless of errors.

PR #372, opened 2023-11-22
feat: allow extra owners in the cleanup controllers

Unix4ever:
If extra owners is passed, the controller will not skip the resources owned by the specified owners and will destroy them too.

### DIFF
--- a/pkg/controller/generic/cleanup/cleanup.go
+++ b/pkg/controller/generic/cleanup/cleanup.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/siderolabs/gen/xerrors"
+	"github.com/siderolabs/gen/xslices"
 	"go.uber.org/zap"
 
 	"github.com/cosi-project/runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


pkg/controller/generic/cleanup/cleanup.go
```diff
@@ -256,11 +257,35 @@ func HasNoOutputs[O generic.ResourceWithRD, I generic.ResourceWithRD](
...
+	extraOwners map[string]struct{}
+}
+
+// RemoveOutputsOption defines the function for passing optional arguments to the RemoveOutputs handler.
+type RemoveOutputsOption func(options *RemoveOutputsOptions)
+
+// WithExtraOwners enables destroy for the resources which have the specified owners.
+func WithExtraOwners(owners ...string) RemoveOutputsOption {
+	return func(options *RemoveOutputsOptions) {
+		options.extraOwners = xslices.ToMap(owners, func(k string) (string, struct{}) {
```
utkuozdemir, line 271: You can use `xslices.ToSet` here

utkuozdemir (approved):
Like the idea

PR #373, opened 2023-11-24
feat: implement pre/post transform hooks in the Transform controller

Unix4ever:
Pre/post methods are called before and after the transform controller runs the loop over all processed resources.


pkg/controller/generic/transform/controller.go
```diff
@@ -240,6 +254,12 @@ func (ctrl *Controller[Input, Output]) Run(ctx context.Context, r controller.Run
 			return err
 		}
 
+		if ctrl.postTransformHook != nil {
```
smira, line 257: should we move it _after_ cleanupOutputs, which might call finalizer hooks?

smira (approved):

pkg/controller/generic/transform/controller.go
Unix4ever: sure

PR #381, opened 2023-12-20
feat: implement QController and qruntime

smira:
Rearrange the existing rruntime adapter: move it to a separate package,
extract interfaces, etc.

Move more internal packages to `internal/` to separate interfaces from
internal details.

QController processes one item from the queue of changed items, so the
reconcile is only triggered for updated items.

In other terms, QController has same set of features as existing
Controllers: metrics, backoff, etc.

New conformance tests for QControllers.

Implement generic/qtransform controller which is (almost) compatible
with generic/transform controller.


pkg/controller/conformance/runtime.go
```diff
@@ -564,3 +564,149 @@ func (suite *RuntimeSuite) TestControllerRuntimeMetrics() {
...
+		SourceNamespace: srcNS,
+		TargetNamespace: targetNS,
+	}))
+
+	suite.Require().NoError(suite.State.Create(suite.ctx, NewIntResource(srcNS, "id1", 1)))
+	suite.Require().NoError(suite.State.Create(suite.ctx, NewIntResource(srcNS, "id2", 2)))
+
+	suite.startRuntime()
+
+	suite.Assert().NoError(retry.Constant(10*time.Second, retry.WithUnits(10*time.Millisecond)).Retry(
```
Unix4ever, line 583: can we use `rtestutils.AssertResources` here?

smira: no, these resources don't support ResourceWithRD

Unix4ever: ok I see

Unix4ever (approved):
:cool:

smira:
I'll merge this PR to get things moving forward before the NY, and open a (smaller) PR with more fixes and stuff, but still looking towards review feedback we can address in a follow-up PR.

PR #391, opened 2024-01-17
fix: pass through correctly requeue errors via qtransform

smira:
The controller should unwrap the error to allow `Modify` call to succeed, but it should re-wrap it back into RequeueError when returning to qruntime.

### DIFF
--- a/pkg/controller/generic/qtransform/qtransform.go
+++ b/pkg/controller/generic/qtransform/qtransform.go
@@ -7,6 +7,7 @@ package qtransform
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/siderolabs/gen/optional"
@@ -225,8 +226,17 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 		}
 	}
 
+	var requeueError *controller.RequeueError
+
 	if err := safe.WriterModify(ctx, r, mappedOut, func(out Output) error {
-		return ctrl.transformFunc(ctx, r, logger, in, out)
+		transformError := ctrl.transformFunc(ctx, r, logger, in, out)
+
+		// unwrap requeue error, so that we don't fail the modify if requeue was done without an explicit error
+		if errors.As(transformError, &requeueError) {
+			return requeueError.Err()
+		}
+
+		return transformError
 	}); err != nil {
 		if state.IsConflictError(err) {
 			// conflict due to wrong phase, skip it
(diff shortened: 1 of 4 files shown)


pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -237,9 +247,17 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 			return nil
 		}
 
+		if requeueError != nil {
```
smira, line 250: I think if `err != nil` here and it's not unwrapped `requeueError`, we should return it, not `requeueError`.

What I mean is: if `Modify` fails e.g. because the resource can't be saved, we should fail hard, not `requeueError`.

smira (commented):
I can't approve my own PR, but :+1:

pkg/controller/generic/qtransform/qtransform.go
utkuozdemir: Hm, don't we already do that?

Above, if `transformError` is not a `RequeueError`, we return it as-is (line 239).

And here, we are in the error case already (because line 240).
If this was indeed a `requeueError`, this line 250 will evaluate to true, and we will return it.

If not, it will return the error as-is on line 254, as you suggested.

Or am I missing something?

smira: I mean the following case (for example):

1. `requeueError` - is a wrapped with `nil` `Err()`
2. so at the line 236 we return `nil`
3. `Modify` fails for its own reasons, so `err != nil`
4. on the line 251 we will return `requeueError` masking the real `err`

utkuozdemir: Ok, I think I got it now - please have another look

pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -237,7 +247,15 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 			return nil
 		}
 
-		return err
+		if requeueError == nil || requeueError.Err() == nil {
```
smira, line 250: this is still incorrect I think. 

```go
if requeueError != nil && requeueError.Err() == err` {
   return requeueError
}

return err
```

utkuozdemir: yes, now I read it and it finally clicked for me :D changed.

pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -237,9 +247,19 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 			return nil
 		}
 
+		if requeueError != nil && requeueError.Err() == err { //nolint:errorlint
+			// if requeueError was specified, and Modify returned in unmodified, return it
```
smira, line 251:
```suggestion
			// if requeueError was specified, and Modify returned it unmodified, return it
```

utkuozdemir (approved):

DmitriyMV (commented):
Note to future self: what happens here is "Unwrap -> Detach Head -> return from Modify -> Attach Head -> Return"

PR #392, opened 2024-01-25
feat: implement controller runtime watch cache

smira:
If enabled, controller runtime uses a side-effect of watching resources to build an in-memory view of resources being read to substitute controller reads from the state with reads from the cache.


pkg/controller/runtime.go
```diff
@@ -113,7 +113,6 @@ type Output struct {
 type Reader interface {
 	Get(context.Context, resource.Pointer, ...state.GetOption) (resource.Resource, error)
 	List(context.Context, resource.Kind, ...state.ListOption) (resource.List, error)
-	WatchFor(context.Context, resource.Pointer, ...state.WatchForConditionFunc) (resource.Resource, error)
```
frezbo, line 116: intended?

pkg/controller/runtime/internal/cache/cache.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+	"fmt"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/controller/runtime/options"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+)
+
+// ResourceCache provides a read-only view of resorces which implements controller.Reader interface.
```
frezbo, line 18:
```suggestion
// ResourceCache provides a read-only view of resources which implements controller.Reader interface.
```

pkg/controller/runtime.go
smira: yes... I believe we're not using it anywhere (need to verify though)

frezbo: i thought it was used in talos, not sure

smira: it is used indeed, but not in controllers, so gc'ed unused thing, and it's confusing.

smira: confirmed that it's not use neither in Talos nor in Omni

pkg/state/impl/inmem/collection.go
```diff
@@ -456,7 +456,13 @@ func (collection *ResourceCollection) WatchAll(ctx context.Context, singleCh cha
 					}
 				}
 
-				if !channel.SendWithContext(ctx, singleCh, state.Event{Type: state.Bootstrapped}) {
+				if !channel.SendWithContext(
+					ctx, singleCh,
+					state.Event{
+						Type:     state.Bootstrapped,
+						Resource: resource.NewTombstone(resource.NewMetadata(collection.ns, collection.typ, "", resource.VersionUndefined)),
```
utkuozdemir, line 463: Why send a tombstone here?

utkuozdemir: Nvm, saw the state-etcd PR which explains it.

utkuozdemir (approved):
🔥

pkg/state/impl/inmem/collection.go
smira: well, the idea is to send which Watch bootstrapped, as if you Watch multiple resource types, and send all watch events to the same channel, there is no way to tell which one boostrapped.

Tombstone is used just as something which carries the minimum Resource metadata (we need to send namespace & type only).

smira:
Updated with even more tests:

* Metrics test in controller runtime now verifies that with cache we see less reads hitting state
* Tests for the `ResourceCache` itself.

PR #401, opened 2024-02-06
feat: allow setting a custom logger for the user operations

utkuozdemir:
Allow passing a `UserLogger` to the runtime, which will be used where the control is passed to the user of the runtime, i.e., the implementor of the controllers.

With this, it is possible to have different log levels/configuration between the runtime internals and the application code that is using it as its framework.

### DIFF
--- a/pkg/controller/generic/cleanup/cleanup.go
+++ b/pkg/controller/generic/cleanup/cleanup.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic"
+	"github.com/cosi-project/runtime/pkg/logging"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
@@ -85,6 +86,10 @@ func (ctrl *Controller[I]) Outputs() []controller.Output {
 
 // Run implements controller.Controller interface.
 func (ctrl *Controller[I]) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	if internalLogger, ok := ctx.Value(logging.InternalLoggerContextKey{}).(*zap.Logger); ok {
+		logger = internalLogger
+	}
+
 	for {
 		select {
 		case <-ctx.Done():
--- a/pkg/controller/generic/cleanup/cleanup_test.go
+++ b/pkg/controller/generic/cleanup/cleanup_test.go
@@ -13,9 +13,11 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/goleak"
+	"go.uber.org/zap"
 
 	"github.com/cosi-project/runtime/pkg/controller/generic/cleanup"
 	"github.com/cosi-project/runtime/pkg/controller/runtime"
+	"github.com/cosi-project/runtime/pkg/controller/runtime/options"
 	"github.com/cosi-project/runtime/pkg/future"
 	"github.com/cosi-project/runtime/pkg/logging"
 	"github.com/cosi-project/runtime/pkg/resource"
@@ -30,9 +32,10 @@ func runTest(t *testing.T, f func(ctx context.Context, t *testing.T, st state.St
 
 	st := state.WrapCore(namespaced.NewState(inmem.Build))
 
-	logger := logging.DefaultLogger()
+	logger := logging.DefaultLogger().With(zap.String("type", "internal"))
+	userLogger := logging.DefaultLogger().With(zap.String("type", "user"))
 
-	rt, err := runtime.NewRuntime(st, logger)
+	rt, err := runtime.NewRuntime(st, logger, options.WithUserLogger(userLogger))
 	require.NoError(t, err)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
--- a/pkg/controller/generic/transform/controller.go
+++ b/pkg/controller/generic/transform/controller.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic"
+	"github.com/cosi-project/runtime/pkg/logging"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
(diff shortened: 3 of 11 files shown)


smira (commented):
I'm not sure I got the idea - if we pass a different logger (which might be just Logger, not UserLogger?) to the controller runtime, why do we need to pass it around more, in the end controller-runtime distributes the logger to the `Run` and `Reconcile` functions.

utkuozdemir:
> I'm not sure I got the idea - if we pass a different logger (which might be just Logger, not UserLogger?) to the controller runtime, why do we need to pass it around more, in the end controller-runtime distributes the logger to the `Run` and `Reconcile` functions.

Basically to for example set `Warn` level for the COSI internals, but to `Info` where the control is passed to a controller in `Run` - i.e. where we implement it.

smira:
I just don't get the idea still - `ctx` is distributed by controller-runtime same way as `logger`, so why are we pulling a logger from `ctx`? if `logger` and `ctx` are passed together?

I don't remember COSI internals logging a lot vs. the controller-runtime though.

utkuozdemir:
> I just don't get the idea still - `ctx` is distributed by controller-runtime same way as `logger`, so why are we pulling a logger from `ctx`? if `logger` and `ctx` are passed together?
> 
> I don't remember COSI internals logging a lot vs. the controller-runtime though.

Ah, the thing about the context is kind of a workaround. So, simply:

- When you create a new controller/transformcontroller/qcontroller, the `logger` argument passed to your implementation will the the so called "user logger", while COSI itself will be logging its things to an internal logger.

- The thing with the context is a kind ow workaround to make the `CleanupController` and `TransformController`s "internal parts" to be able to grab the internal logger and log to it, instead of the user logger (because the one that is passed to `Run` now is the user one). For example, for these kind of logs to go to the internal logger:
   ```
   cleanup: waiting for resources to be destroyed
   transform: triggered teardown of output resource
   ```

So, from the controller implementor perspective, there are no changes - they don't need to do anything with the context etc, they just get passed the right logger.

smira:
ok, that's good, but I don't see the idea still, I'm probably missing something here.

The logger passed by the controller-runtime is enhanced with fields related to the controller, the item being processed, etc. 

if we want to have a filter on what is logged and what is not, it feels strange to me that we introduce two different loggers, I think the proper way is to use the logging level and (possibly) additional filters based on what we can pass through the loggers.

We could use a 'factory' pattern to generate loggers for specific controllers (by name) if want to customize behavior (but I doubt it makes a lot of sense), but having two loggers for all controllers seems strange to me.

So I mean the dimension of separating what COSI logs and what controller logs seems less useful then e.g. the dimension of what controller A logs vs. controller B. But I doubt even that makes a lot of sense.

If we see some logs being less useful/confusing, we can drop/adjust the level.

utkuozdemir:
> ok, that's good, but I don't see the idea still, I'm probably missing something here.

I think I need to explain my train of thought in detail here:
- Our app can use libraries A, B and C. They are black boxes for our app, and they log things internally with the level that makes sense for them.
- The "log level" can, and often does have different meanings between the library itself and the application that is using it. For example, it makes perfect sense for COSI to log lots of internal stuff it does as `info` - for COSI, they are `info`. But it does not mean that for Omni they are `info`s as well - they might as well be "noise", or "debug" from Omni perspective.

  In fact, what you do [here](https://github.com/siderolabs/omni/blob/269deb3a89e49311dd4452b889a84e3a47d5682a/internal/backend/runtime/omni/state_etcd.go#L143-L159) is exactly an example of that - what is `debug` level for etcd is not even `debug` for us - therefore, you pass a custom logger to it. So even when we use `debug` level in our app, we won't get spammed with the etcd logs.
  Similarly, what I do [here](https://github.com/siderolabs/omni/pull/1569/files#diff-ed32333589a322f7884f076f67519a8510a162489b961198014dea4f9cd4b601R79) is another example of that - what is `info` for the Kubernetes library might not be `info` from our perspective.
- COSI, partly because it is a _framework_, passes control to our code here and there (`Run` etc.). So, we can say that there are (at least) two types of logs produced by the COSI - its internal logs, and the logs where it passes the control to the implementor (because it passes a logger).
- We might want to have different levels and/or other config for these 2 types of logs.

> The logger passed by the controller-runtime is enhanced with fields related to the controller, the item being processed, etc.

This is also the answer of why give an additional logger to COSI: Because as you said, it enriches the logger before passing the control to the user - with the controller name, etc. So, essentially, when I give COSI a `UserLogger`, I am telling it "take this logger, and when you pass control to me, enrich it and pass it back to me as well. In other words, we don't lose the enrichment this way.

Btw, about the name `UserLogger` - it might not be the perfect name, but it's just naming. We could instead call the regular `logger` and the existing logger `internalLogger` - the idea stays the same.

> if we want to have a filter on what is logged and what is not, it feels strange to me that we introduce two different loggers, I think the proper way is to use the logging level and (possibly) additional filters based on what we can pass through the loggers.

Adjustments to the level in COSI imo is the right answer here, mainly because of the first 2 bullet points above.

> We could use a 'factory' pattern to generate loggers for specific controllers (by name) if want to customize behavior (but I doubt it makes a lot of sense), but having two loggers for all controllers seems strange to me.
> So I mean the dimension of separating what COSI logs and what controller logs seems less useful then e.g. the dimension of what controller A logs vs. controller B. But I doubt even that makes a lot of sense.

Per-controller level control would be a nice addition, but I think this and that are orthogonal, not mutually exclusive. To be clear, the problem I had was not this or that logger specifically - it was simply, COSI runtime, as a whole, printed too much of its internal logs that I might not need. Not a per-controller thing.

> If we see some logs being less useful/confusing, we can drop/adjust the level.

I know I already explained it, but just to put it in a different way one more time: I don't think adjusting the level is the right approach. It could work for COSI, yes, because it's our code - but try seeing it as yet another library (like the etcd lib) - we wouldn't just go and change its internal log levels. We, instead, would pass it a logger with the level we like.

Finally, please note that this does not mean that we are reducing any log level anywhere or changing anything in the log behavior: when nothing is passed, user logger is set to be the same as the internal logger, so everything works as-is. This just allows us to have finer-grained control, if we want, where we want.

smira:
I got your point, but my train of thought is try to not overcomplicate things if we can, and I think it is a perfect example that we own COSI, so we can make it work the way we want without overcomplications.

I'd guess making successful "map" reconcile events from qruntime be Debug would solve 90% of the problems in a single line change.

utkuozdemir:
> I got your point, but my train of thought is try to not overcomplicate things if we can, and I think it is a perfect example that we own COSI, so we can make it work the way we want without overcomplications.
> 
> I'd guess making successful "map" reconcile events from qruntime be Debug would solve 90% of the problems in a single line change.

Ok, I don't have anything to add here. Don't think it would do any harm, and compared to the existing complexity I don't think this added much, but you have the final say. Feel free to close.

PR #403, opened 2024-02-09
feat: add a new utility method `qtransform.WithExtraInput`

utkuozdemir:
Allow specifying the input type for extra qcontroller inputs, e.g., `DestroyReady`.

### DIFF
--- a/pkg/controller/generic/qtransform/options.go
+++ b/pkg/controller/generic/qtransform/options.go
@@ -68,10 +68,14 @@ type ControllerOptions struct {
 // ControllerOption is an option for QTransformController.
 type ControllerOption func(*ControllerOptions)
 
-// WithExtraMappedInput adds an extra  mapped inputs to the controller.
-func WithExtraMappedInput[I generic.ResourceWithRD](mapFunc MapperFuncGeneric[I]) ControllerOption {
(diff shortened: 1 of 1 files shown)


pkg/controller/generic/qtransform/options.go
```diff
@@ -68,11 +68,15 @@ type ControllerOptions struct {
 // ControllerOption is an option for QTransformController.
 type ControllerOption func(*ControllerOptions)
 
-// WithExtraMappedInput adds an extra  mapped inputs to the controller.
-func WithExtraMappedInput[I generic.ResourceWithRD](mapFunc MapperFuncGeneric[I]) ControllerOption {
+// WithExtraInput adds an extra input with the given kind to the controller.
+func WithExtraInput[I generic.ResourceWithRD](mapFunc MapperFuncGeneric[I], inputKind controller.InputKind) ControllerOption {
```
smira, line 72: idk, probably WithExtraMappedInputKind (?) as it's still Mapped, but custom kind

smira (approved):

pkg/controller/generic/qtransform/options.go
utkuozdemir: done

PR #407, opened 2024-02-20
fix: make `RequeueError` never push existing items execution time

Unix4ever:
If the controller returns `RequeueError` it should not schedule another update for a later time if there already is a task in the queue to be executed earlier.

### DIFF
--- a/pkg/controller/runtime/internal/qruntime/internal/containers/priority_queue.go
+++ b/pkg/controller/runtime/internal/qruntime/internal/containers/priority_queue.go
@@ -26,7 +26,7 @@ type PriorityQueue[T comparable] struct {
 // Push item to the queue with releaseAfter time.
 //
 // If the item is not in the queue, it will be added.
-// If the item is in the queue, and releaseAfter is greater than the existing releaseAfter, it will be re-added in the new position.
+// If the item is in the queue, and releaseAfter is less than the existing releaseAfter, it will be re-added in the new position.
 //
 // Push returns true if the item was added, and false if the existing item in the queue was updated (or skipped).
 func (queue *PriorityQueue[T]) Push(item T, releaseAfter time.Time) bool {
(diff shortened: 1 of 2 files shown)


pkg/controller/runtime/internal/qruntime/internal/containers/priority_queue.go
```diff
@@ -26,7 +26,7 @@ type PriorityQueue[T comparable] struct {
 // Push item to the queue with releaseAfter time.
 //
 // If the item is not in the queue, it will be added.
-// If the item is in the queue, and releaseAfter is greater than the existing releaseAfter, it will be re-added in the new position.
+// If the item is in the queue, and releaseAfter is greater than the existing releaseAfter, it will be not be added to the queue.
```
smira, line 29: this comment looks to be off now, as it now re-inserts with smaller timestamp, and tests should be adjusted

pkg/controller/runtime/internal/qruntime/internal/containers/priority_queue_test.go
```diff
@@ -110,12 +110,12 @@ func TestPriorityQueueDeduplicateSimple(t *testing.T) {
 func TestPriorityQueueDeduplicateWithReleaseAfter(t *testing.T) {
 	var q containers.PriorityQueue[int]
 
+	var base time.Time
+
 	for _, i := range []int{1, 2, 3, 4, 5} {
-		assert.True(t, q.Push(i, time.Time{}))
+		assert.True(t, q.Push(i, base.Add(time.Duration(20-i))))
```
smira, line 116: these are different elements, something is broken if this needs to change

pkg/controller/runtime/internal/qruntime/internal/containers/priority_queue.go
smira: and we need tests for new behavior, the current thing doesn't look good

pkg/controller/runtime/internal/qruntime/internal/containers/priority_queue_test.go
Unix4ever: I add some items with time greater than the time in the next loop. That's the test for the new logic.
But I've actually expanded the tests now, so that it checks both adding the items with greater timestamps and smaller ones.

pkg/controller/runtime/internal/qruntime/internal/containers/priority_queue.go
Unix4ever: Updated

smira (approved):

PR #416, opened 2024-02-28
chore: rekres, update Go to 1.22

smira:
Update Go modules, fix linting errors.


.drone.yml
```diff
@@ -68,11 +68,11 @@ steps:
   depends_on:
   - base
 
-- name: unit-tests-race
+- name: unit-tests%!(EXTRA string=-race)
```
frezbo, line 71: is this new?

frezbo (approved):

.drone.yml
smira: I think this is a bug in kres, we don't use `drone.yml` here, so it doesn't matter

PR #417, opened 2024-02-28
feat: implement resource self-destroy controller

Unix4ever:
For resources not owned by any controller allow using the controller that will destroy resources which have no finalizers.


pkg/controller/generic/destroy/destroy.go
```diff
@@ -0,0 +1,98 @@
...
+	if in.Metadata().Owner() != "" {
+		return nil
+	}
+
+	// do not do anything while the resource has any finalizers
+	if !in.Metadata().Finalizers().Empty() {
+		return nil
+	}
+
+	logger.Info("destroy resource without finalizers", zap.String("id", in.Metadata().ID()))
```
smira, line 90: `id` should be already in the `zap` context

smira (approved):

PR #423, opened 2024-03-15
feat: use rangefunc for iterators

utkuozdemir:
Give rangefunc a try.

### DIFF
--- a/pkg/safe/state.go
+++ b/pkg/safe/state.go
@@ -7,10 +7,11 @@ package safe
 import (
 	"context"
 	"fmt"
+	"iter" //nolint:gci
 	"slices"
 
 	"github.com/siderolabs/gen/channel"
-	"github.com/siderolabs/gen/xslices"
+	"github.com/siderolabs/gen/xslices" //nolint:gci
 
 	"github.com/cosi-project/runtime/pkg/controller/generic"
 	"github.com/cosi-project/runtime/pkg/resource"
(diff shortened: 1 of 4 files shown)


DmitriyMV:
Depends on https://github.com/dominikh/go-tools/issues/1494

pkg/safe/state.go
```diff
@@ -283,10 +284,27 @@ func (l *List[T]) Find(fn func(T) bool) (T, bool) {
...
+
+			if !yield(res) {
+				return
+			}
+		}
+	}
+}
+
 // ListIterator is a generic iterator over resource.Resource slice.
 type ListIterator[T any] struct {
```
DmitriyMV, line 309: Needs deprecation notice.

pkg/safe/state.go
```diff
@@ -283,10 +284,27 @@ func (l *List[T]) Find(fn func(T) bool) (T, bool) {
...
 
 // Iterator returns a new iterator over the list.
+//
+// Deprecated: use [List.Iter] instead.
 func (l *List[T]) Iterator() ListIterator[T] {
 	return ListIterator[T]{pos: 0, list: *l}
 }
 
+// Iter returns a new rangefunc iterator over the list.
+func (l *List[T]) Iter() iter.Seq[T] {
```
DmitriyMV, line 294: All because if this https://pkg.go.dev/iter#hdr-Naming_Conventions

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cosi-project/runtime
 
-go 1.22.0
+go 1.22.1
```
DmitriyMV, line 3:
```suggestion
go 1.23.0
```

pkg/safe/state.go
utkuozdemir: can do, but also feel free to take this over if you want to push this faster forward.

DmitriyMV: Okay, I'll take it.

utkuozdemir: 🙏

DmitriyMV (approved):

PR #452, opened 2024-05-31
feat: add an error tag to destroy outputs

utkuozdemir:
Add a new tag called `DestroyOutputTag` that is similar to `SkipReconcileTag`, which destroys the output resource.

### DIFF
--- a/pkg/controller/generic/qtransform/qtransform.go
+++ b/pkg/controller/generic/qtransform/qtransform.go
@@ -21,10 +21,17 @@ import (
 	"github.com/cosi-project/runtime/pkg/state"
 )
 
+var errPendingOutputTeardown = errors.New("output is being torn down")
+
 // SkipReconcileTag is used to tag errors when reconciliation should be skipped without an error.
 //
 // It's useful when next reconcile event should bring things into order.
 type SkipReconcileTag struct{}
+
+// DestroyOutputTag is used to tag errors when output should be destroyed without an error.
+//
+// It's useful when output is not needed anymore.
+type DestroyOutputTag struct{}
 
 // QController provides a generic implementation of a QController which implements controller transforming Input resources into Output resources.
 //
@@ -226,6 +233,14 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 		}
 	}
 
+	if err := ctrl.handleOutputTearingDown(ctx, r, mappedOut); err != nil {
+		if errors.Is(err, errPendingOutputTeardown) {
+			return nil
+		}
+
+		return err
+	}
+
 	var requeueError *controller.RequeueError
 
 	if err := safe.WriterModify(ctx, r, mappedOut, func(out Output) error {
@@ -247,6 +262,10 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 			return nil
 		}
 
+		if xerrors.TagIs[DestroyOutputTag](err) {
+			return ctrl.handleDestroyOutput(ctx, r, mappedOut)
+		}
+
 		if requeueError != nil && requeueError.Err() == err { //nolint:errorlint
 			// if requeueError was specified, and Modify returned it unmodified, return it
 			// otherwise Modify failed for its own reasons, and use that error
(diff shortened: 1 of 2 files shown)


Unix4ever (approved):
:cool:

pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -247,6 +252,14 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 			return nil
 		}
 
+		if xerrors.TagIs[DestroyOutputTag](err) {
```
smira, line 255: what about tear down and finalizers on the the output resource? I think we should restructure the flow here (idk how) to move into `reconcileTearingDown` phase... otherwise the finalizers on the input are still set, it looks a bit strange?

utkuozdemir: idk either, let's them hold this, it anyway is only a quality-of-life improvement.

utkuozdemir: can you have another look, did some changes after our discussion.

pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -263,6 +281,52 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
...
+		return false, nil
+	}
+
+	if output.Metadata().Phase() != resource.PhaseTearingDown {
+		return false, nil
+	}
+
+	// output is being torn down
+
+	if !mappedOut.Metadata().Finalizers().Empty() { // not ready to be destroyed yet
```
smira, line 303: should be `output`?

pkg/controller/generic/qtransform/qtransform_test.go
```diff
@@ -621,7 +679,7 @@ func setup(t *testing.T, f func(ctx context.Context, st state.State, rt *runtime
 	rt, err := runtime.NewRuntime(st, logger)
 	require.NoError(t, err)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Hour)
```
smira, line 682: :exploding_head:

pkg/controller/generic/qtransform/qtransform_test.go
```diff
@@ -278,6 +283,59 @@ func TestMapWithErrors(t *testing.T) {
...
+		rtestutils.AssertResource[*B](ctx, t, st, "transformed-1", func(r *B, assert *assert.Assertions) {
+			assert.Equal(resource.PhaseTearingDown, r.Metadata().Phase())
+			assert.Equal(`"regular"-0`, r.TypedSpec().Out)
+		})
+
+		// remove the finalizer on the output
+		require.NoError(t, st.RemoveFinalizer(ctx, NewB("transformed-1", BSpec{}).Metadata(), "some-finalizer"))
+
+		// the output should be destroyed now
+		rtestutils.AssertNoResource[*B](ctx, t, st, "transformed-1")
```
smira, line 335: we should verify there that new resource got created?

utkuozdemir: ah, I missed that part. Changed again, now I think it works as it should, but not sure about what I do to get it to requeue here: https://github.com/cosi-project/runtime/pull/452/files#diff-ad6500f6caec4671f4830d1c934e07a96c03689c01e0a65a1565ec2e0a279ffaR301

pkg/controller/generic/qtransform/qtransform_test.go
utkuozdemir: forgot to revert 🤦

pkg/controller/generic/qtransform/qtransform.go
utkuozdemir: good catch, fixed.

pkg/controller/generic/qtransform/qtransform_test.go
smira: I think in that case you need to return an error which will return before calling `WriterModify`, but it won't be an error from a controller point of view if finalizers are not empty

pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -226,6 +232,10 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 		}
 	}
 
+	if err := ctrl.handleOutputTearingDown(ctx, r, mappedOut); err != nil {
+		return err
```
smira, line 236:
```suggestion
	if errors.Is(err, errNotReadyPendingFinalizersSkipReconciliation) {
	    return nil
	}
	
	return err
```

pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -263,6 +277,46 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
...
+	if err != nil && !state.IsNotFoundError(err) {
+		return err
+	}
+
+	if output == nil || output.Metadata().Phase() != resource.PhaseTearingDown {
+		return nil
+	}
+
+	// output is being torn down, destroy it if it is ready to be destroyed
+	if output.Metadata().Finalizers().Empty() {
```
smira, line 294:
```suggestion
	if !output.Metadata().Finalizers().Empty() {
	    return errNotReadyPendingFinalizersSkipReconciliation
	}
	
	/// destroy
	
	
	return nil /// continue to WriterModify
```

pkg/controller/generic/qtransform/qtransform.go
```diff
@@ -263,6 +277,46 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
...
+		if err = r.Destroy(ctx, mappedOut.Metadata()); err != nil && !state.IsNotFoundError(err) {
```
smira, line 295:
```suggestion
		if err = r.Destroy(ctx, mappedOut.Metadata()); err != nil  {
```

pkg/controller/generic/qtransform/qtransform_test.go
utkuozdemir: done as discussed

PR #465, opened 2024-06-19
chore: bump deps and remove false lock

DmitriyMV:
Bump those things:
- actions/checkout to v4.1.2
- codecov/codecov-action to v4.1.0
- github.com/klauspost/compress v1.17.8
- google.golang.org/protobuf v1.34.1

And remove lock around `runtime.(*Runtime).group.Wait` call.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/gertd/go-pluralize v0.2.1
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.20.0
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/klauspost/compress v1.17.8
+	github.com/klauspost/compress v1.17.9
 	github.com/planetscale/vtprotobuf v0.6.0
 	github.com/siderolabs/gen v0.5.0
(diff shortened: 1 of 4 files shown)


utkuozdemir (approved):

Unix4ever (approved):

go.mod
```diff
@@ -46,3 +46,5 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20240513163218-0867130af1f8 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240513163218-0867130af1f8 // indirect
 )
+
+retract v0.4.7 // Wait with locked mutex leads to the deadlock
```
frezbo, line 50: what does this do?

DmitriyMV: I don't think I can explain better that [Go](https://go.dev/ref/mod#go-mod-file-retract) devs [can](https://play-with-go.dev/retract-module-versions_go119_en/). But basic idea here is to say that this version should't be used and explain the reason. The tooling knows about this directive and will skip this version.

PR #470, opened 2024-07-04
feat: export controller-runtime cached state

smira:
Allow the consumer to take advantage of the cached state for Get/List operations.


frezbo:
would this affect talos? like it gets cached data?

smira:
> would this affect talos? like it gets cached data?

no, this is completely optional, just exposed the internal cache of the controller-runtime as state.State, it doesn't change anything if explicitly being used

Unix4ever (approved):

smira:
Added some unit-tests

PR #472, opened 2024-07-05
chore: add two functions to the `safe` package

DmitriyMV:
• Add `Swap` method to the `safe.List` type, so it can be sorted in place. 
• Add `ToSlice` function which converts `safe.List` to slice using the provided function.


Unix4ever (approved):
❤️

PR #474, opened 2024-07-16
feat: add safe.List.SortFunc

smira:
It should make it way easier to sort resource lists.

### DIFF
--- a/pkg/safe/state.go
+++ b/pkg/safe/state.go
@@ -7,6 +7,7 @@ package safe
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


DmitriyMV:
Maybe we can reuse ideas from SortBonds in Talos?

smira:
> Maybe we can reuse ideas from SortBonds in Talos?

yes, that's why I implemented this, as `SortBonds` is a nightmare, too much code to just sort the resources

DmitriyMV (approved):

PR #475, opened 2024-07-18
fix: ignore output not found errors in the cleanup controller

Unix4ever:
We shouldn't care if the resource doesn't exist if we're trying to destroy it.

### DIFF
--- a/pkg/controller/generic/cleanup/cleanup.go
+++ b/pkg/controller/generic/cleanup/cleanup.go
@@ -324,7 +324,7 @@ func RemoveOutputs[O generic.ResourceWithRD, I generic.ResourceWithRD](
 				}
 
 				ready, err := r.Teardown(ctx, out.Metadata(), controller.WithOwner(owner))
-				if err != nil {
+				if err != nil && !state.IsNotFoundError(err) {
 					multiErr = multierror.Append(multiErr, fmt.Errorf("error tearing down %q resource %q: %w", zeroOutputType, out.Metadata().ID(), err))
 
 					continue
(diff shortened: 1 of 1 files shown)


utkuozdemir (approved):

pkg/controller/generic/cleanup/cleanup.go
```diff
@@ -325,7 +325,9 @@ func RemoveOutputs[O generic.ResourceWithRD, I generic.ResourceWithRD](
 
 				ready, err := r.Teardown(ctx, out.Metadata(), controller.WithOwner(owner))
 				if err != nil {
-					multiErr = multierror.Append(multiErr, fmt.Errorf("error tearing down %q resource %q: %w", zeroOutputType, out.Metadata().ID(), err))
+					if !state.IsNotFoundError(err) {
```
smira, line 328: You could merge into a single if to reduce nesting?

pkg/controller/generic/cleanup/cleanup.go
```diff
@@ -338,7 +340,9 @@ func RemoveOutputs[O generic.ResourceWithRD, I generic.ResourceWithRD](
 
 				err = r.Destroy(ctx, out.Metadata(), controller.WithOwner(owner))
 				if err != nil {
-					multiErr = multierror.Append(multiErr, fmt.Errorf("error destroying %q resource %q: %w", zeroOutputType, out.Metadata().ID(), err))
+					if !state.IsNotFoundError(err) {
```
smira, line 343: And same here?

smira (approved):

PR #492, opened 2024-09-11
fix: skip update only if the conflict error is related to main output

Unix4ever:
Fix `qtransform` controller bug which was skipping the reconciliation if the conflict error is returned from the `transformFunc` and is related to the extra outputs.

Introduce a way to check if the error was raised by any particular resource type.


Unix4ever:
> I guess we need to apply same fix to state-etcd

Yeah, I'm going to do it,

PR #494, opened 2024-09-17
fix: abort watch correctly for protobuf client

smira:
There were two bugs:

* `io.EOF` was ignored for no good reason
* the aggregated/non-aggregated was not properly handled for errors

This should fix COSI controller runtime not crashing on connection close via gRPC.


pkg/state/protobuf/client/client.go
```diff
@@ -403,21 +404,30 @@ func (adapter *Adapter) WatchKindAggregated(ctx context.Context, resourceKind re
...
 			},
-		)
+			)
+		}
 	}
 
 	for {
 		msg, err := cli.Recv()
+		if err != nil {
+			log.Printf("watchAdapter: %v", err)
```
frezbo, line 429: needed?

frezbo (approved):

pkg/state/protobuf/client/client.go
smira: :hankey:  thank you, I always forget to remove debugging

Unix4ever (approved):

PR #511, opened 2024-11-05
test: fix a data race in the test

smira:
Don't share resource Metadata() between watch goroutine and tests.

### DIFF
--- a/pkg/state/conformance/state.go
+++ b/pkg/state/conformance/state.go
@@ -980,7 +980,7 @@ func (suite *StateSuite) testWatchKindWithBookmarks(useAggregated bool) {
 	initial, err := suite.State.List(ctx, res.Metadata())
 	suite.Require().NoError(err)
 
-	suite.Require().NoError(watchAggregateAdapter(ctx, useAggregated, suite.State, res.Metadata(), ch, state.WithBootstrapContents(true)))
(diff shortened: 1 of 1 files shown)


utkuozdemir (approved):

frezbo (approved):

pkg/state/conformance/state.go
```diff
@@ -980,7 +980,7 @@ func (suite *StateSuite) testWatchKindWithBookmarks(useAggregated bool) {
 	initial, err := suite.State.List(ctx, res.Metadata())
 	suite.Require().NoError(err)
 
-	suite.Require().NoError(watchAggregateAdapter(ctx, useAggregated, suite.State, res.Metadata(), ch, state.WithBootstrapContents(true)))
+	suite.Require().NoError(watchAggregateAdapter(ctx, useAggregated, suite.State, res.Metadata().Copy(), ch, state.WithBootstrapContents(true)))
```
DmitriyMV, line 983:
```suggestion
	suite.Require().NoError(watchAggregateAdapter(ctx, useAggregated, suite.State, *res.Metadata(), ch, state.WithBootstrapContents(true)))
```

This should suffice no?

smira: not clear, as Metadata() has some mutable fields, Copy is more straightforward

PR #512, opened 2024-11-05
fix: make inmem bookmarks random for each run

smira:
As inmem storage is not persisted, make sure watch bookmark from one run doesn't work with another run.

This still allows watches to be restarted on connection failures, but if the watch is restarted on a program restart, bookmark won't match anymore.

### DIFF
--- a/pkg/state/errors.go
+++ b/pkg/state/errors.go
@@ -137,3 +137,15 @@ func errPhaseConflict(r resource.Reference, expectedPhase resource.Phase) error 
 		},
 	}
 }
+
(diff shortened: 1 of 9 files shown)


frezbo (approved):

smira:
TODO for myself:

* make watch bookmark error special (so that `errors.Is` can check for it)
* make watch retry skip retries on fatal watch bookmark errors

Unix4ever (approved):

smira:
Verified with Talos Linux:

```
talosctl get cpustat -w
```

* successfully retries transient errors (e.g. `apid` restarts) and keeps watching
* on a machine reboot, watch correctly aborts (as the bookmark becomes invalid)

PR #539, opened 2024-12-19
fix: fix `kv` equality check for empty values

utkuozdemir:
We did not distinguish between an entry with an empty value vs a non-existent key. This caused resource labels/annotations to not get updated in some cases.


pkg/resource/internal/kv/kv.go
```diff
@@ -92,7 +92,8 @@ func (kv KV) Equal(other KV) bool {
 	}
 
 	for k, v := range kv.m {
-		if v != other.m[k] {
+		otherV, ok := other.m[k]
+		if !ok || v != otherV {
```
DmitriyMV, line 96: Single line maybe?

DmitriyMV (approved):

pkg/resource/internal/kv/kv.go
```diff
@@ -92,7 +92,8 @@ func (kv KV) Equal(other KV) bool {
 	}
 
 	for k, v := range kv.m {
-		if v != other.m[k] {
+		otherV, ok := other.m[k]
```
DmitriyMV, line 95: Just drop the whole block and use `maps. Equal`

smira (approved):
nice catch!

pkg/resource/internal/kv/kv.go
utkuozdemir: hehe yes that actually makes sense :D

PR #543, opened 2024-12-27
feat: implement correct watch restart for controller runtime

### LINKED ISSUE #540: Controller-Runtime running using a state over GRPC via `StateClient` crashes when it loses connection to the server
Observed it when using https://github.com/siderolabs/omni + https://github.com/siderolabs/omni-infra-provider-bare-metal.

The moment I stop Omni  (ran via `make docker-compose-up WITH_DEBUG=true`) by CTRL+C, the provider's controller runtime crashes with:
```
Error: failed to run provider: failed to run components: controller runtime watch error: rpc error: code = Internal desc = server closed the stream without sending trailers
2024/12/19 14:05:27 failed to run: failed to run provider: failed to run components: controller runtime watch error: rpc error: code = Internal desc = server closed the stream without sending trailers
```

Let me know if more info is needed.

smira:
Controller runtime does `WatchKind` on all controller input resources, but initial bookmark value is only delivered if there is any update received (on each of the watches).

If resource cache is used, `BootstrapContents` will deliver the bookmark, but if the cache is not used, and there are no updates, no bookmark is ever sent, so watch can't be restarted.

Fix this by adding yet another option to send 'Noop' events which carry only the bookmark allowing watch retry code to handle it.

Fixes #540


smira:
Uses https://github.com/cosi-project/specification/pull/14

I will do a similar change to `state-etcd` to support initial bookmarks.

frezbo (approved):

utkuozdemir (approved):

PR #550, opened 2025-01-17
fix: don't report that we retry context canceled

smira:
It suppresses logging and returns a bit earlier (mostly cosmetic fix).

### DIFF
--- a/pkg/state/protobuf/client/client.go
+++ b/pkg/state/protobuf/client/client.go
@@ -462,7 +462,7 @@ func (adapter *Adapter) WatchKindAggregated(ctx context.Context, resourceKind re
 	return nil
 }
 
-//nolint:gocognit,gocyclo,cyclop
+//nolint:gocognit,gocyclo,cyclop,maintidx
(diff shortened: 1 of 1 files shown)


pkg/state/protobuf/client/client.go
```diff
@@ -513,6 +513,14 @@ func (adapter *Adapter) watchAdapter(
 		for {
 			// retry loop - at the beginning of the loop 'err' is the error to be retried,
 			// lastBookmark is the last seen bookmark
+			//
+			// quick check for context canceled, no need to retry
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+			}
```
DmitriyMV, line 522:
```suggestion
			if err := ctx.Err(); err != nil {
				return nil, err
			}
```

smira: thanks, updated

frezbo (approved):

Unix4ever (approved):

PR #552, opened 2025-01-22
fix: do not overwrite error in watch retry

utkuozdemir:
Do not overwrite the watch error with the error from the context when checking for the context error, so that if we hit the maximum retry attempts, the error will be printed correctly.


smira (approved):
thank you! my bad

Unix4ever (approved):

PR #567, opened 2025-02-21
test: ignore phase conflict errors in the `rtestutils.Destory`

Unix4ever:
Teardown might fail with phase conflict if the `Teardown` was called in another process.
Just ignore that error, it means that teardown was already called.


Unix4ever:
> do we need a test?

This scenarios will be really hard to test. As it happens due to pretty rare race.

frezbo (approved):

PR #574, opened 2025-03-03
fix: invalid error message in safe.List functions

Orzelius:
I've been playing around with cosi the past few days and it's been a lovely experience, but I ran into a misleading error message and decided to figure out the inner workings and fix it myself. Here's the fix with some additional tests to catch similar stuff in the future.

**Changes**

* Add test cases for the type validation logic
* Fix the error message for when state returns a wrong type in `safe.StateList` and `safe.ReaderList` functions

Before this fix a message like this would have been returned when running the newly added tests
```
type mismatch on the first element: 
expected *conformance.Resource[string,conformance.strSpec,*conformance.strSpec], 
got resource.List
```
* Add and use `typeMismatchFirstElErr` helper
* Use already existing `typeMismatchErr`


pkg/safe/state_test.go
```diff
@@ -145,3 +148,105 @@ func TestListFilter(t *testing.T) {
...
+			return safe.StateGetByID[*testResource](context.Background(), s, "")
+		},
+		"StateList": func(_ *testing.T) (any, error) {
+			return safe.StateList[*testResource](context.Background(), s, resource.Metadata())
+		},
+		"StateListAll": func(_ *testing.T) (any, error) {
+			return safe.StateListAll[*testResource](context.Background(), s)
+		},
+		"StateWatch": func(t *testing.T) (any, error) {
+			assert.NoError(t, safe.StateWatch[*testResource](context.Background(), s, resource.Metadata(), channel))
```
Orzelius, line 222: my code editor is complaining about unnecessary type arguments (can be inferred from args), but there doesn't seem to be a linter that mandates me to remove them and they can be found in other places as well so I decided to keep the types for simplicity. Please let me know if they should be removed.

smira:
lgtm, thanks, requesting another review from @DmitriyMV

pkg/safe/state_test.go
```diff
@@ -145,3 +148,105 @@ func TestListFilter(t *testing.T) {
...
+
+func TestTypeValidation(t *testing.T) {
+	s := &brokenState{}
+	resource := testResource{}
+	channel := make(chan safe.WrappedStateEvent[*testResource])
+
+	type testCase = func(t *testing.T) (result any, err error)
+
+	cases := map[string]testCase{
+		"StateGet": func(_ *testing.T) (any, error) {
```
DmitriyMV, line 209:
```suggestion
		"StateGet": func(*testing.T) (any, error) {
```

Here and below.

pkg/safe/state_test.go
DmitriyMV: The usual rule is that if type argument can be inferred it should be inferred.

pkg/safe/state_test.go
```diff
@@ -145,3 +148,105 @@ func TestListFilter(t *testing.T) {
...
+			return safe.StateListAll[*testResource](context.Background(), s)
+		},
+		"StateWatch": func(t *testing.T) (any, error) {
+			assert.NoError(t, safe.StateWatch[*testResource](context.Background(), s, resource.Metadata(), channel))
+
+			select {
+			case event := <-channel:
+				return event.Resource()
+			case <-time.After(time.Second):
+				panic("timed out waiting for event")
```
DmitriyMV, line 228: No reason to panic, as we have access to `t.Fatal` here.

pkg/safe/state_test.go
```diff
@@ -145,3 +148,105 @@ func TestListFilter(t *testing.T) {
...
+			}
+		},
+		"StateWatchKind": func(t *testing.T) (any, error) {
+			assert.NoError(t, safe.StateWatchKind[*testResource](context.Background(), s, resource.Metadata(), channel))
+
+			select {
+			case event := <-channel:
+				return event.Resource()
+			case <-time.After(time.Second):
+				panic("timed out waiting for event")
```
DmitriyMV, line 238: No reason to panic, as we have access to t.Fatal here.

DmitriyMV (approved):

pkg/safe/state_test.go
Orzelius: Ah, good point! addressed in [79ca968](https://github.com/cosi-project/runtime/pull/574/commits/79ca968c4c93d77238a19a57189a4eb849b4f3fd)

pkg/safe/state_test.go
Orzelius: Changing to `t.Fatal` would require to return something after the for block to satisfy the function so I decided that a panic would be cleaner.

Orzelius:
@DmitriyMV Thank you for the feedback! I've address the comments in [79ca968](https://github.com/cosi-project/runtime/pull/574/commits/79ca968c4c93d77238a19a57189a4eb849b4f3fd).

pkg/safe/state_test.go
DmitriyMV: Yeah, but it will fail the entire test-suite and not just specific test.

PR #575, opened 2025-03-05
feat: add support for ModifyOptions for controllers

smira:
This allows to customize expected phase when modifying resources.


DmitriyMV (approved):

smira:
Tested this with Talos.

PR #588, opened 2025-04-16
feat: allow aliases to be skipped in resource definitions

smira:
This allows to "hide" from `get` commands some resources unless they are called by the full name.

See https://github.com/siderolabs/talos/issues/10666

Also rekres and bump deps, fix linting issues.


pkg/resource/meta/spec/resource_definition.go
```diff
@@ -82,29 +84,31 @@ func (spec *ResourceDefinitionSpec) Fill() error {
 	}
 
 	spec.DisplayType = pluralizeClient.Singular(name)
-	spec.Aliases = append(spec.Aliases, strings.ToLower(spec.DisplayType))
 
-	spec.AllAliases = append(spec.AllAliases, strings.ToLower(name))
+	if !spec.SkipAutomaticAliases {
```
smira, line 88: this is the actual change 🤦‍♂️

frezbo (approved):

PR #593, opened 2025-05-06
feat: support non blocking mode in `TeardownAndDestroy` state method

Unix4ever:
This mode is useful in the controllers. They can call this method repeatedly.


Unix4ever:
actually hold on, I need to fix the controllers interfaces too.

smira:
> This mode is useful in the controllers. They can call this method repeatedly.

In controllers, we can do another version of this method if we want to which doesn't block. I think the `State` was good int he sense that it's an easy shortcut if you're okay to block.

Unix4ever:
Closing this PR, will first add the helpers for teardown to the Omni repo.
Will verify the concept there and then add them to the upstream.

PR #601, opened 2025-05-16
refactor: split parts of the adapter as owned.State

smira:
The idea is to split the part which enforces the ownership (but skip things like rate-limiting protection and controller-specific access checks) into a separate package which has exactly same interface.

This allows one to easily build a `controller.Reader/Writer` out of `state.State` (e.g. in the unit-tests).


smira:
- [ ] need to add some tests for `owned.State`

Unix4ever (approved):

PR #617, opened 2025-07-04
feat: add optional resource assertion

shanduur:
Add new AssertOptionalResources assertion function, that checks if at
least one of the provided resources exists.

### DIFF
--- a/pkg/resource/rtestutils/assertions.go
+++ b/pkg/resource/rtestutils/assertions.go
@@ -289,6 +289,95 @@ func AssertLength[R ResourceWithRD](ctx context.Context, t *testing.T, st state.
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/resource/rtestutils/assertions.go
```diff
@@ -289,6 +289,95 @@ func AssertLength[R ResourceWithRD](ctx context.Context, t *testing.T, st state.
 	}
 }
 
+// AssertOptionalResources asserts that at least one of the resources exists and passes the assertionFunc.
+func AssertOptionalResources[R ResourceWithRD](
```
smira, line 293: nit: better naming?

```suggestion
func AssertOneOfResources[R ResourceWithRD](
```


also I wonder if we could make it simply an option to `AssertResources` with "how many 'passed' is enough?"

PR #636, opened 2025-09-01
fix: correct deduplication of watch events

smira:
This is extracted from #633 as a separate fix.

The idea is that if we merge/deduplicate watch events, we should do this by immutable attributes of the resource: namespace, type, id.

The previous implementation might split two events on e.g. finalizers being empty, so two events which report finalizers empty/finalizers non-empty might be delivered in the wrong order: first empty, next non-empty.

It might not trigger a real bug due to the way controllers work, but still is not consistent and correct.

### DIFF
--- a/pkg/controller/runtime/internal/reduced/reduced.go
+++ b/pkg/controller/runtime/internal/reduced/reduced.go
@@ -7,11 +7,26 @@ package reduced
 
 import "github.com/cosi-project/runtime/pkg/resource"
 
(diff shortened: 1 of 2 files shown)


Unix4ever (approved):

frezbo (approved):
no tests?

smira:
> no tests?

yes, existing tests test this indirectly, it's the very core part of COSI

utkuozdemir (approved):

PR #663, opened 2026-05-06
refactor: remove type parameter from destroy controller

utkuozdemir:
The destroy controller does not need its input type parameter. The type was only used to recover the resource definition at construction time and to read the resource during reconcile, neither of which require static typing. Switch to a value-based constructor that takes a resource definition, and keep the typed constructor as a thin wrapper for back compatibility.

This allows registering destroy controllers in a loop over a list of resource types, instead of having to enumerate them at compile time.

### DIFF
--- a/pkg/controller/generic/destroy/destroy.go
+++ b/pkg/controller/generic/destroy/destroy.go
@@ -15,45 +15,50 @@ import (
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic"
 	"github.com/cosi-project/runtime/pkg/resource"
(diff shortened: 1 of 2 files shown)


Unix4ever (approved):
Looks good to me

smira (commented):
unrelated note, but I wonder this controller name is a valid metric value in Prometheus with `[]`

smira (approved):

utkuozdemir:
> unrelated note, but I wonder this controller name is a valid metric value in Prometheus with `[]`

Good point, I had a look, since we also have more of that, like `LabelsExtractor[%s]` etc. in Omni - they are only used as label values, and they seem to accept any UTF-8, so we are good.

PR #664, opened 2026-05-07
feat: log and meter reconcile skips in QController runtime

oguzkilcan:
Previously qtransform swallowed `SkipReconcileTag` errors and returned nil, so the QRuntime saw a plain success and the reason behind the skip was lost. The runtime now detects the tag itself, emits a dedicated `reconcile skipped` log line carrying the original reason, and increments a new `qcontroller_skips` expvar metric — while still treating the cycle like a success for backoff purposes.

### DIFF
--- a/pkg/controller/generic/qtransform/qtransform.go
+++ b/pkg/controller/generic/qtransform/qtransform.go
@@ -268,7 +268,7 @@ func (ctrl *QController[Input, Output]) reconcileRunning(ctx context.Context, lo
 		}
 
 		if xerrors.TagIs[SkipReconcileTag](err) {
-			return nil
+			return err
 		}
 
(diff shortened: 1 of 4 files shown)


smira (approved):

utkuozdemir (approved):

utkuozdemir:
> @oguzkilcan merge was not performed: push: To https://github.com/cosi-project/runtime.git
>  ! [remote rejected] main -> main (cannot lock ref 'refs/heads/main': is at fc693cb31d313d7ee573461b798006a994fc559f but expected 282c9a0522a64a85074b59133a99d27db799e10b)
> error: failed to push some refs to 'https://github.com/cosi-project/runtime.git'
> 

I got this as well

PR #665, opened 2026-05-11
test: use correct context to avoid race in TestContextWithTeardown

oguzkilcan:
Use correct context to avoid race in TestContextWithTeardown


pkg/state/wrap.go
```diff
@@ -7,12 +7,21 @@ package state
...
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/siderolabs/go-pointer"
 
 	"github.com/cosi-project/runtime/pkg/resource"
 )
 
+// contextWithTeardownCancelGrace is how long ContextWithTeardown waits between
```
smira: I'm not a big fan of the delay like that... do we have a real-life example of this problem?

smira (approved):

PR #668, opened 2026-07-01
fix: properly handle finalizers empty watch in TeardownAndDestroy call

Unix4ever:
`TeardownAndDestroy` hangs if `Teardown` and `Destroy` is called concurrently from two different actors.
Just `WatchFor` with `WithFinalizersEmpty` doesn't catch the moment when the resource is destroyed by another actor. Replace it with another fine carved method for that.
Also add reproducer tests.

### DIFF
--- a/pkg/state/teardown_and_destroyer_test.go
+++ b/pkg/state/teardown_and_destroyer_test.go
@@ -7,11 +7,13 @@ package state_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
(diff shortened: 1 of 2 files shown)


pkg/state/teardown_and_destroyer_test.go
```diff
@@ -120,6 +120,77 @@ func TestCoreWrapperTeardownAndDestroyFallback(t *testing.T) {
 	assert.True(t, state.IsNotFoundError(err))
 }
 
+// racyDestroyCoreState wraps a CoreState and runs a hook the first time Watch
+// is called, simulating another party mutating/destroying the resource exactly
+// as coreWrapper.TeardownAndDestroy starts watching for finalizers to drain.
+type racyDestroyCoreState struct {
+	state.CoreState
+	onWatch func()
+	fired   bool
```
shanduur:
```suggestion
	fired   atomic.Bool
```

pkg/state/teardown_and_destroyer_test.go
```diff
@@ -120,6 +120,77 @@ func TestCoreWrapperTeardownAndDestroyFallback(t *testing.T) {
...
+}
+
+func (r *racyDestroyCoreState) Watch(ctx context.Context, ptr resource.Pointer, ch chan<- state.Event, opts ...state.WatchOption) error {
+	if r.onWatch != nil && !r.fired {
+		r.fired = true
+
+		r.onWatch()
+	}
```
shanduur:
```suggestion
	if r.onWatch != nil && r.fired.CompareAndSwap(false, true) {
		r.onWatch()
	}
```

oguzkilcan (approved):

shanduur:
I don't think this needs to be racy, right?

pkg/state/teardown_and_destroyer_test.go
```diff
@@ -120,6 +120,77 @@ func TestCoreWrapperTeardownAndDestroyFallback(t *testing.T) {
...
+	// Another party drains the finalizer and destroys the resource exactly as we
+	// begin watching, so the initial (and only) watch event is Destroyed.
+	racy.onWatch = func() {
+		require.NoError(t, st.RemoveFinalizer(ctx, r.Metadata(), "fin"))
+		require.NoError(t, st.Destroy(ctx, r.Metadata()))
+	}
+
+	done := make(chan error, 1)
+
+	go func() {
```
smira: can we just run `t.Run()` multiple times with `t.Parallel()`, all of them doing TeardownAndDestroy()? they should hit the race, no?

smira: at least I think this matches our production bug?

Unix4ever: that also works, yeah

smira (approved):
nice catch!

pkg/state/teardown_and_destroyer_test.go
Unix4ever: that was deleted

PR #674, opened 2026-08-04
feat: rework inmem resource state implementation

smira:
This bring inmem closer to other state implementations:

* `inmem.NewState()` is now namespaced inside, not an instance per namespace
* one inmem state instance has a shared event buffer (stream) for all resource types and namespaces

This should provide better memory usage for any inmem user, and also it builds the foundation to build a shared, consistent watch API for the controller runtime (not in this PR).


pkg/state/impl/inmem/internal/eventbuffer/eventbuffer.go
```diff
@@ -0,0 +1,413 @@
...
+// resumeLocked returns the number of the feed events covered by the bookmark.
+//
+// The buffer mutex should be held.
+func (f *Feed) resumeLocked(bookmark state.Bookmark) (int64, error) {
+	seq, err := decodeBookmark(bookmark)
+	if err != nil {
+		return 0, err
+	}
+
+	if seq < f.staleSeq || seq > f.publishedSeq {
```
shanduur, line 321: Bookmarks encode only the process cookie and a feed-local sequence number. This range check therefore accepts a bookmark from another namespace/type whenever that sequence happens to fall inside this feed's valid range, and the resumed watcher silently skips that many target-feed events.

Reproducer: save this beside `eventbuffer_test.go` and run:

```console
go test ./pkg/state/impl/inmem/internal/eventbuffer -run TestReviewCrossFeedBookmarkMustBeRejected -count=1
```

```go
func TestReviewCrossFeedBookmarkMustBeRejected(t *testing.T) {
    buf := eventbuffer.New(16, 16, 1)
    source := buf.NewFeed("source-ns", "source-type")
    target := buf.NewFeed("target-ns", "target-type")

    source.Publish(testEvent("source-1"))
    sourceWatcher, err := source.Subscribe(eventbuffer.SubscribeOptions{TailEvents: 1})
    require.NoError(t, err)
    sourceEvents, err := sourceWatcher.Next(t.Context())
    require.NoError(t, err)
    sourceWatcher.Close()
    require.Len(t, sourceEvents, 1)

    target.Publish(testEvent("target-1"))
    target.Publish(testEvent("target-2"))

    _, err = target.Subscribe(eventbuffer.SubscribeOptions{Bookmark: sourceEvents[0].Bookmark})
    require.Error(t, err, "a bookmark from another feed must not be accepted")
}
```

At this head it fails with `An error is expected but got nil`. Please bind bookmarks to a feed identity, including buffer identity and namespace/type, validate it here, and retain this as a regression test.

pkg/state/impl/inmem/inmem.go
```diff
@@ -14,61 +14,77 @@ import (
...
+
+	if r, ok := st.collections.Load(key); ok {
 		return r
 	}
 
-	collection := NewResourceCollection(st.ns, typ, st.initialCapacity, st.maxCapacity, st.gap, st.store)
-
-	r, _ := st.collections.LoadOrStore(typ, collection)
+	// a nil BackingStore converts to a nil collection.Store, so the collection stays in-memory only
+	r, _ := st.collections.LoadOrStore(key, collection.New(ns, typ, st.buffer, st.store))
```
shanduur, line 80: `collection.New` is evaluated before `LoadOrStore`, and it immediately registers a feed in `buf.feeds`. Concurrent first access to the same `(namespace,type)` can therefore register several feeds while only one collection wins the map insertion. The losing feeds remain reachable forever and every cleanup sweep scans them under the global buffer mutex.

Reproducer: save this as `pkg/state/impl/inmem/concurrent_collection_review_test.go` and run:

```console
go test ./pkg/state/impl/inmem -run TestReviewConcurrentCollectionCreationDoesNotLeakFeeds -count=1
```

```go
package inmem

import (
    "reflect"
    "sync"
    "testing"
)

func TestReviewConcurrentCollectionCreationDoesNotLeakFeeds(t *testing.T) {
    for attempt := 0; attempt < 100; attempt++ {
        st := NewState()
        start := make(chan struct{})
        var wg sync.WaitGroup

        for range 128 {
            wg.Add(1)
            go func() {
                defer wg.Done()
                <-start
                st.getCollection("same-ns", "same-type")
            }()
        }

        close(start)
        wg.Wait()

        feeds := reflect.ValueOf(st.buffer).Elem().FieldByName("feeds").Len()
        if feeds != 1 {
            t.Fatalf("attempt %d: one installed collection registered %d feeds", attempt, feeds)
        }
    }
}
```

On my run it failed immediately with `attempt 0: one installed collection registered 4 feeds`. Please make creation single-winner or unregister losing candidates, and retain a concurrent same-key regression test.

pkg/state/impl/inmem/internal/collection/watch.go
```diff
@@ -0,0 +1,328 @@
...
+
+	r := slc[:0]
+
+	for _, v := range slc {
+		if fn(&v) {
+			r = append(r, v)
+		}
+	}
+
+	return r
```
shanduur, line 327: Please clear the discarded tail before returning, for example with `clear(slc[len(r):])`. The compacted slice shares its backing array with the original event batch, so an aggregated-watch consumer retaining a small filtered batch also retains all rejected `state.Event` resource references beyond `len(r)`. That defeats history cleanup for those resources and can retain a full backlog for the lifetime of the consumer-held batch.

shanduur (changes requested):
I found two correctness/resource-lifecycle issues that should be fixed before merge. Full locally executed reproducers for the first two findings are included inline.

pkg/state/impl/inmem/inmem.go
```diff
@@ -14,61 +14,77 @@ import (
...
-			gap:             options.HistoryGap,
-		}
+	st := &State{
+		collections: concurrent.NewHashTrieMap[collectionKey, *collection.Collection](),
+		buffer:      eventbuffer.New(options.HistoryInitialCapacity, options.HistoryMaxCapacity, options.HistoryGap),
+		store:       options.BackingStore,
+	}
+
+	if options.HistoryCleanupCtx != nil && options.HistoryCleanupInterval > 0 {
+		go st.buffer.RunCleanup(options.HistoryCleanupCtx, options.HistoryCleanupInterval)
```
utkuozdemir, line 66: Wondering if 
- we should worry about this goroutine panicking
- leave concurrency to the caller? Like a blocking `RunCleanup(ctx) error` or similar?

pkg/state/impl/inmem/internal/eventbuffer/eventbuffer.go
smira: I'm not sure if it's fully valid though - this is internal package, so it might not be reachable in general as this implies someone used bookmark from one Watch call in the other unrelated Watch. Same way one could in theory supply any other bookmark.

pkg/state/impl/inmem/inmem.go
smira: this is great catch!

pkg/state/impl/inmem/internal/collection/watch.go
smira: this is pre-existing (this function got moved around), but nice catch!

pkg/state/impl/inmem/inmem.go
smira: if we do this, I think we should move the history cleanup interval as well out

pkg/state/impl/inmem/internal/eventbuffer/cleanup.go
```diff
@@ -0,0 +1,111 @@
...
+		buffered := &buf.stream[pos%capacity]
+		pos = buffered.prevPos
+
+		// only the event is released: the feed is what expireLocked accounts the slot to once it
+		// is finally overwritten, and prevPos keeps the chain of the feed walkable
+		buffered.event = state.Event{}
+	}
+
+	// a new watch can no longer start from the dropped events
+	f.staleSeq = boundary
```
utkuozdemir, line 93: The cleanup and the expiry account the same events into `staleSeq` twice. The sweep assigns it here after releasing the events, and when the write position later wraps over those same slots, `expireLocked` increments it once more per slot. For a quiet feed `staleSeq` can then exceed `publishedSeq`, which makes the tail rewind in `Subscribe` negative, so a new watcher starts ahead of the feed and the next published events are silently skipped instead of being delivered or reported as an overrun. Also, once the buffer wraps, `boundary <= f.staleSeq` becomes permanently true and the sweep stops releasing anything.

Reproducer: save this beside `eventbuffer_test.go` and run:

```console
go test ./pkg/state/impl/inmem/internal/eventbuffer -run TestReviewCleanupMustNotSkipFutureEvents -count=1
```

```go
func TestReviewCleanupMustNotSkipFutureEvents(t *testing.T) {
	buf := eventbuffer.New(8, 8, 1)
	quiet := buf.NewFeed("ns", "quiet")
	busy := buf.NewFeed("ns", "busy")

	quiet.Publish(testEvent("old"))

	// the feed has no watchers, so two sweeps release the "old" event
	require.Zero(t, buf.Cleanup())
	require.Equal(t, 1, buf.Cleanup())

	// rotate the buffer over the released slot
	for i := range 8 {
		busy.Publish(testEvent(strconv.Itoa(i)))
	}

	tail, err := quiet.Subscribe(eventbuffer.SubscribeOptions{TailEvents: 1})
	require.NoError(t, err)
	defer tail.Close()

	quiet.Publish(testEvent("new"))

	ctx, cancel := context.WithTimeout(t.Context(), 100*time.Millisecond)
	defer cancel()

	events, err := tail.Next(ctx)
	require.NoError(t, err)
	require.Equal(t, []resource.ID{"new"}, eventIDs(events))
}
```

At this head it fails with the watcher receiving no events at all, the "new" event published to the feed is never delivered.

pkg/state/impl/inmem/internal/eventbuffer/eventbuffer.go
```diff
@@ -0,0 +1,413 @@
...
+		buf.stream[buf.evictedUpTo%capacity].feed.evictedSeq++
+	}
+
+	// the events within the gap from the write position are about to be overwritten soon,
+	// so a new watch should not start from them
+	//
+	// the target is raised to evictedUpTo to keep staleSeq >= evictedSeq with a zero gap, and it is
+	// clamped to writePos, as the events past it are not written yet (this only matters if the gap
+	// is (mis)configured to be bigger than the capacity)
+	staleTarget := min(max(buf.writePos-capacity+int64(buf.gap), buf.evictedUpTo), buf.writePos)
```
utkuozdemir, line 229: While the buffer is still growing, no event can be overwritten yet (the capacity doubles exactly at `writePos == capacity`, before any slot is reused), so the gap protects nothing at that stage. But the stale boundary here is computed from the current capacity, and `staleSeq` only ever grows, so events get permanently marked stale while the buffer is nowhere near full. On `main` this window was computed at subscribe time against the already grown capacity, so growth made the early events reachable again. The same scenario through the public API replays all events on `main` and loses some of them at this head.

Two ways this shows up: with the default options (initial capacity 256, gap 50) a fresh state loses tail and bookmark access to its earliest events after about 200 publishes. With an initial capacity below the gap (e.g. `WithHistoryInitialCapacity(8)` with the default gap) almost every event goes stale the moment it is published, and bookmarks handed out one event ago are already rejected.

Reproducer covering both, save it beside `eventbuffer_test.go` and run:

```console
go test ./pkg/state/impl/inmem/internal/eventbuffer -run TestReviewGrowthMustNotMakeEventsStale -count=1
```

```go
func TestReviewGrowthMustNotMakeEventsStale(t *testing.T) {
	for _, test := range []struct {
		name            string
		initialCapacity int
		gap             int
		publish         int
	}{
		{name: "gap below initial capacity", initialCapacity: 16, gap: 8, publish: 20},
		{name: "gap above initial capacity", initialCapacity: 8, gap: 50, publish: 6},
	} {
		t.Run(test.name, func(t *testing.T) {
			buf := eventbuffer.New(test.initialCapacity, 40960, test.gap)
			feed := buf.NewFeed("ns", "type")

			for i := range test.publish {
				feed.Publish(testEvent(strconv.Itoa(i)))
			}

			w, err := feed.Subscribe(eventbuffer.SubscribeOptions{TailEvents: test.publish})
			require.NoError(t, err)

			defer w.Close()

			events, err := w.Next(t.Context())
			require.NoError(t, err)
			require.Len(t, events, test.publish, "no event was ever overwritten, all should be replayable")
		})
	}
}
```

At this head both subtests fail, replaying 13 of 20 and 1 of 6 events.

pkg/state/impl/store/bolt/example_test.go
```diff
@@ -73,24 +73,23 @@ func Example() {
...
-	))
+	// * in-memory: any other namespace, e.g. runtime
+	resources := state.WrapCore(namespaced.NewState(func(ns resource.Namespace) state.CoreState {
+		switch ns {
+		case "persistent", "system":
+			return persistentState
+		default:
+			return inmem.NewState()
+		}
+	}))
```
utkuozdemir, line 92: I think we should create the ephemeral inmem state here only once, with the new changes. Otherwise each ephemeral namespace gets its own state and event buffer.

```suggestion
	// an inmem.State is either fully backed by the store, or fully ephemeral, so build a separate
	// persistent state, and route the namespaces to it with namespaced.NewState
	persistentState := inmem.NewStateWithOptions(
		inmem.WithBackingStore(backingStore),
	)

	// a single ephemeral state instance is shared by all the other namespaces
	ephemeralState := inmem.NewState()

	// create resource state with following namespaces
	// * backed by BoltDB: persistent, system
	// * in-memory: any other namespace, e.g. runtime
	resources := state.WrapCore(namespaced.NewState(func(ns resource.Namespace) state.CoreState {
		switch ns {
		case "persistent", "system":
			return persistentState
		default:
			return ephemeralState
		}
	}))
```

pkg/state/impl/inmem/options.go
```diff
@@ -79,10 +119,13 @@ func WithBackingStore(store BackingStore) StateOption {
...
+// As the history buffer is shared by all namespaces and resource types of the state, the default
+// capacity is much bigger than the depth which used to be reserved per resource type.
 func DefaultStateOptions() StateOptions {
 	return StateOptions{
-		HistoryMaxCapacity:     100,
-		HistoryInitialCapacity: 100,
-		HistoryGap:             5,
+		HistoryMaxCapacity:     40960,
+		HistoryInitialCapacity: 256,
+		HistoryGap:             50,
```
utkuozdemir, line 129: Should we update the recommendation on `WithHistoryGap` ("gap 5% of the capacity")? The new defaults (gap 50 for max capacity 40960) are way below 5%.

pkg/state/impl/inmem/internal/eventbuffer/eventbuffer.go
```diff
@@ -0,0 +1,413 @@
...
+	}
+}
+
+// collectLocked returns the last count events of the feed in the publish order.
+//
+// The buffer mutex should be held, and the events being collected should still be in the buffer.
+func (f *Feed) collectLocked(count int64) []state.Event {
+	buf := f.buf
+
+	events := make([]state.Event, count)
```
utkuozdemir, line 394: Wondering if, for the single-resource watches (by ID), we could do the filtering here in `collectLocked`, e.g. by applying `SubscribeOptions.Filter`, which already exists for the tail rewind. This way the whole event batch of the type would not be copied to each per-ID watcher on every wakeup, only to be dropped in the ID check afterwards.

pkg/state/impl/inmem/inmem.go
utkuozdemir: I'm torn, so your call - I often lean towards that kind of design myself, also remember us having discussions about us not liking libraries spawning their goroutines.

pkg/state/impl/inmem/internal/eventbuffer/eventbuffer.go
```diff
@@ -0,0 +1,450 @@
...
+		return
+	}
+
+	// the events within the gap from the write position are about to be overwritten soon,
+	// so a new watch should not start from them
+	//
+	// the target is raised to evictedUpTo to keep staleSeq >= evictedSeq with a zero gap, and it is
+	// clamped to writePos, as the events past it are not written yet (this only matters if the gap
+	// is (mis)configured to be bigger than the capacity)
+	staleTarget := min(max(buf.writePos-capacity+int64(buf.gap), buf.evictedUpTo), buf.writePos)
```
utkuozdemir, line 257: The stale boundary is computed from `writePos` before this publish, but a watcher subscribing right after it sees the buffer with one more event. So once the buffer is full, a tail watch can start `gap - 1` slots away from the eviction edge instead of `gap`, and with gap 1 it gets overrun by the very next publish. On `main` the same scenario kept the gap.

Reproducer, save beside `eventbuffer_test.go`:

```go
func TestReviewTailKeepsTheGap(t *testing.T) {
	buf := eventbuffer.New(4, 4, 1)
	f := buf.NewFeed("ns", "type")

	for i := range 4 {
		f.Publish(testEvent(strconv.Itoa(i)))
	}

	w, err := f.Subscribe(eventbuffer.SubscribeOptions{TailEvents: 4})
	require.NoError(t, err)
	t.Cleanup(w.Close)

	f.Publish(testEvent("4"))

	events, err := w.Next(t.Context())
	require.NoError(t, err)
	require.Equal(t, []resource.ID{"1", "2", "3", "4"}, eventIDs(events))
}
```

At this head it fails with `buffer overrun: ... published 5, evicted 1, consumed 0`.

pkg/state/impl/inmem/internal/eventbuffer/cleanup.go
```diff
@@ -0,0 +1,117 @@
...
+	// the events published as of the previous sweep are old enough, and the events every watcher
+	// has consumed are of no use to the watchers which are already subscribed
+	boundary := min(f.sweepSeq, f.consumedLocked())
+
+	f.sweepSeq = f.publishedSeq
+
+	// the events at or below the stale boundary were either dropped by an earlier sweep or pushed
+	// out of the buffer altogether, and the ones below evictedSeq are not even ours to read anymore,
+	// which is covered as staleSeq >= evictedSeq
+	staleBoundary := f.staleBoundaryLocked()
```
utkuozdemir, line 66: The sweep releases only the events above `max(staleSeq, droppedSeq)`, so the events which are stale because of the gap but still sitting in the buffer are never released by the cleanup. If the publishing stops, they keep their resources pinned for good, up to `gap` events (50 with the defaults).

Reproducer, save beside `eventbuffer_test.go`:

```go
func TestReviewCleanupReleasesStaleEvents(t *testing.T) {
	buf := eventbuffer.New(4, 4, 2)
	f := buf.NewFeed("ns", "type")

	for i := range 4 {
		f.Publish(testEvent(strconv.Itoa(i)))
	}

	require.Zero(t, buf.Cleanup())
	require.Equal(t, 4, buf.Cleanup())
}
```

At this head the second sweep returns 3.

pkg/state/impl/namespaced/namespaced.go
```diff
@@ -39,7 +43,20 @@ func (st *State) getNamespace(ns resource.Namespace) state.CoreState { //nolint:
...
+	// with something shared, or run a goroutine of its own)
+	st.namespacesMu.Lock()
+	defer st.namespacesMu.Unlock()
+
+	// re-check after lock
+	if s, ok := st.namespaces.Load(ns); ok {
+		return s
+	}
+
+	s := st.builder(ns)
```
utkuozdemir, line 57: A bit hypothetical, but calling the builder under the lock can deadlock if a builder ever reaches back into the same `namespaced.State` (e.g. reads another namespace), and a slow builder blocks the first access to all the other namespaces. Maybe use a `sync.OnceValue` per namespace instead and drop the lock? The builder still runs once per namespace, only the callers of that namespace wait, and nothing runs under the router's lock:

```go
build, _ := st.namespaces.LoadOrStore(ns, sync.OnceValue(func() state.CoreState {
	return st.builder(ns)
}))

return build()
```

pkg/state/impl/inmem/internal/eventbuffer/eventbuffer.go
utkuozdemir: This question still stands

